PR #13, opened 2019-01-23
fix: instead of using `command` use `which` in Makefiles

UndeadKernel:
The shell built-in `command` used in Makefiles would sometimes not work,
yielding the error `command: Command not found`. This was due to an
optimization that make does: For simple commands, make may avoid
spawning a new shell. `command`, being a built-in shell command, would
then not be found.

The program `which` does not suffer from this problem.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,6 @@
 EMACS ?= emacs
 SHELL = bash
+
 CASK ?= $(shell command -v cask)
 ifeq ($(CASK),)
     $(error "Install cask (https://github.com/cask/cask)")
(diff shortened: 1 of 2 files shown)


dzop:
I'm wondering if we can just provide the `.POSIX` target at the top of the file to avoid this optimization? I'd like to use `command` since it is a shell built-in and avoids dependency on an external program. It even has its own POSIX standard https://pubs.opengroup.org/onlinepubs/009695399/utilities/command.html.

UndeadKernel:
After further investigation, I see the problem because I'm using ZSH instead of Bash.
The following line at the beginning of both files fixes the issue:

```
SHELL=bash -e -o pipefail
```

dzop:
Thanks.

PR #17, opened 2019-01-30
feature: Specify mime precedence using ":display" in org source blocks

UndeadKernel:
A BEGIN_SRC block can specify which mime type to display by enabling the
user to manually set the priority order of mime types.

e.g.:
In the following, :text/plain will be used before considering :text/html
\#BEGIN_SRC jupy-python :display plain html

### DIFF
--- a/jupyter-org-client.el
+++ b/jupyter-org-client.el
@@ -784,7 +784,7 @@ data and metadata of the current MIME type."
   (cl-assert plist json-plist)
   (let* ((params (jupyter-org-request-block-params req))
          (display-mime-types (jupyter-org--find-mime-types
-                              (alist-get :display params nil))))
+                              (alist-get :display params))))
     (when (jupyter-org-request-file req)
       (push (cons :file (jupyter-org-request-file req)) params))
     (cl-destructuring-bind (data metadata)
@@ -1018,23 +1018,24 @@ is a stream result. Otherwise return nil."
 If a match is not found, return nil. Try to be intelligent and
 return what the user might intend to use.
 
-REQ-TYPES can be a strings such as `plain', `plain html' or
+REQ-TYPES can be a string such as `plain', `plain html', or
 `text/plain'. The string `text' is translated to `:text/plain'
 and `image' to `:image/png'."
-  (and req-types ;; proceed only if REQ-TYPES is not nil
-       ;; iterate every specified REQ-TYPE looking for matches in MIME-TYPES
-       (mapcar (lambda (req-type)
-                 (cond
-                  ((not req-type) nil)
-                  ((string= req-type "text") :text/plain)
-                  ((string= req-type "image") :image/png)
-                  (t (let ((regexp (if (string-match "/" req-type)
-                                       req-type
-                                     (concat "/" req-type "$"))))
-                       (cl-loop for ii in jupyter-org-mime-types
-                                if (string-match regexp (symbol-name ii))
-                                return ii)))))
-               (split-string req-types))))
+  (when req-types
+    ;; Iterate the user-specified mimetypes looking for symbols that match a
+    ;; symbol in `jupyter-org-mime-types'. Invalid mimetypes are ignored.
+    (delete nil
+            (mapcar (lambda (req-type)
+                      (cond
+                       ((string= req-type "text") :text/plain)
+                       ((string= req-type "image") :image/png)
+                       ((stringp req-type) (let ((regexp (if (string-match "/" req-type)
+                                                             req-type
+                                                           (concat "/" req-type "$"))))
+                                             (cl-loop for ii in jupyter-org-mime-types
+                                                      if (string-match regexp (symbol-name ii))
+                                                      return ii)))))
+                    (split-string req-types)))))
 
 ;;;;; Fixed width -> example block promotion
 
--- a/jupyter-org-client.el
+++ b/jupyter-org-client.el
@@ -782,13 +782,16 @@ which the above call returns a non-nil value. PARAMS is the REQ's
 `jupyter-org-request-block-params', DATA and METADATA are the
 data and metadata of the current MIME type."
   (cl-assert plist json-plist)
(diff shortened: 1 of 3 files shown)


dzop:
Thanks for this.

Some changes:

1. `jupyter-org-mime-types` should always be used since those are the mimetypes that can currently be displayed by the `org-mode` client.

2. Could you run `indent-region` on all of the changed parts so that the code is properly indented. The indentation doesn't look right in the changed parts of `jupyter-org-result`.

3. In `jupyter-org--find-mime-types` there is no need for the `if` statement since `mapcar` will return nil when operating on an empty list. Then you can just remove the `let` statement as well.

About the implementation:

1. I'm wondering if we can do a little better in finding mimetypes. If a user specified `p` as the value of `:display` it would match both `image/png` and `text/plain`. Maybe having the regex be 
`(concat "/" req-type "$")` so that a full subtype like `png` or `plain` has to be specified?

2. The `mime-types` argument in `jupyter-org--find-mime-types` can be removed since we only need to search the mimetypes in `jupyter-org-mime-types` and this can be done directly inside the function.

UndeadKernel:
Thanks for the input @dzop 
I'm back from vacations and will work on fixing this.

stardiviner:
@UndeadKernel Hi, this is a good idea. And could you add declaration in README too? That'll be helpful for user.

UndeadKernel:
Hi @dzop, I've rebased the changes onto the current master.
I have also added a section to the README file with the new functionality.

Let me know if you see anything that needs changing.

jupyter-org-client.el
```diff
@@ -987,6 +990,30 @@ is a stream result. Otherwise return nil."
...
+
+If a match is not found, return nil. Try to be intelligent and
+return what the user might intend to use.
+
+REQ-TYPES can be a strings such as `plain', `plain html' or
+`text/plain'. The string `text' is translated to `:text/plain'
+and `image' to `:image/png'."
+  (and req-types ;; proceed only if REQ-TYPES is not nil
+       ;; iterate every specified REQ-TYPE looking for matches in MIME-TYPES
+       (mapcar (lambda (req-type)
```
dzop: This function can either return `nil`, a list like `(:text/plain :image/png)`, or a list like `(:text/plain :image/png nil)`. It doesn't look like the last form with `nil` in the list is being considered anywhere though.

If a user puts in an invalid mimetype in the `:display` property there will be a `nil` in the returned list. Maybe we should verify that the `:display` property is valid in `org-babel-execute:jupyter`? I don't think it would be a good idea to do any error handling in this function since this is called after the code block has been evaluated.

stardiviner:
There is one Org results type `:results link` is missing. Might consider support it too?

jupyter-org-client.el
```diff
@@ -987,6 +990,30 @@ is a stream result. Otherwise return nil."
     (when (looking-at-p "\\(?::[\t ]\\|#\\+END_EXAMPLE\\)")
       (line-end-position (unless (eq (char-after) ?:) 0)))))
 
+(defun jupyter-org--find-mime-types (req-types)
```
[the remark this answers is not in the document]
UndeadKernel: I'm new to writing tests using ERT. Let me know if the test I wrote is sufficient.

jupyter-org-client.el
UndeadKernel: The does not return now "nil" if the mimetype is not known. Instead, it is just ignored.

UndeadKernel:
> There is one Org results type `:results link` is missing. Might consider support it too?

Hi @stardiviner, it'd be better if you create a new issue to track this request of yours.

jupyter-org-client.el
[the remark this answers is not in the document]
dzop: 👍

jupyter-org-client.el
```diff
@@ -987,6 +990,32 @@ is a stream result. Otherwise return nil."
...
+If a match is not found, return nil. Try to be intelligent and
+return what the user might intend to use.
+
+REQ-TYPES can be a string such as `plain', `plain html', or
+`text/plain'. The string `text' is translated to `:text/plain'
+and `image' to `:image/png'."
+  (when req-types
+    ;; iterate the user-specified mimetypes looking for symbols that match a
+    ;; symbol in `jupyter-org-mime-types'
+    (remove nil ; remove invalid mimetypes
```
dzop: Can you use `delete` here since `remove` will copy the list which is unnecessary.

Also can you just mention that if the "user-specified" mimetype is not valid it is ignored in the above comment instead of making a new comment.

jupyter-org-client.el
```diff
@@ -987,6 +990,32 @@ is a stream result. Otherwise return nil."
...
+            (mapcar (lambda (req-type)
+                      (cond
+                       ((not req-type) nil)
+                       ((string= req-type "text") :text/plain)
+                       ((string= req-type "image") :image/png)
+                       (t (let ((regexp (if (string-match "/" req-type)
```
dzop: I think it would be better to do `(stringp req-type)` here and then remove the `(not req-type)` clause. `cond` will return nil when none of its clauses succeed anyways.

test/jupyter-test.el
```diff
@@ -260,6 +260,16 @@
           (jupyter-insert :text/html "<p>hello</p>")
           (should html-parser-called))))))
 
+(ert-deftest jupyter-find-mime-types ()
```
dzop: Can you place the test with all the other `org` related tests at the bottom of the file. Also can you name the test the same as the function name, `jupyter-org--find-mime-types`. ERT tests are not normal functions so there is no name clashing.

dzop:
@stardiviner it would be a good idea to support `link`, but I do think that is a separate issue from this pull request.

UndeadKernel:
Thanks for your input and comments @dzop.
Do keep them coming if you find things to improve. There's no better way to learn.

dzop:
Can you rebase this on the lastest master and then this is good to go 👍.

PR #21, opened 2019-02-07
Add variable to change output to tooltip instead of message

sebastianpech:
I added a custom variable to allow usage of tooltips instead of `message` for short messages.
image
For longer return values the output is still forwarded to a buffer.
I implemented it with a customizable variable `jupyter-display-output-function` that is a symbol (either `'message` or `'tooltip`) which is then checked in a `cond` to ensure that only those two cases are used and no arbitrary function can be added.

I hope my code somewhat matches the style of this package

### DIFF
--- a/jupyter-client.el
+++ b/jupyter-client.el
@@ -38,15 +38,23 @@
 (require 'jupyter-messages)
 
 (defcustom jupyter-display-short-eval-result-function #'message
-  "Function used for displaying the result of evaluation which is
-less than 10 lines long. The variable defaults to `message' which
-outputs to the minibuffer.
+  "Function for displaying short evaluation results.
+Evaluation results are cosidered short in case they are less than
+`jupyter-display-short-eval-result-max-lines' lines long.
+The variable defaults to `message' which outputs to the minibuffer.
 
 Any function can be used which takes a string as first argument.
 For example to display the result in a tooltip the variable can
 be set to `popup-tip' from `popup'"
   :group 'jupyter-client
   :type 'function)
+
+(defcustom jupyter-display-short-eval-result-max-lines 10
+  "Maximum number of lines for evaluation results to still be considered short.
+Short evaluation results are displayed using `jupyter-display-short-eval-result-function'.
+Longer results are forwarded to a separate buffer."
+  :group 'jupyter-client
+  :type 'integer)
 
 (declare-function company-begin-backend "ext:company" (backend &optional callback))
 (declare-function company-doc-buffer "ext:company" (&optional string))
@@ -965,7 +973,7 @@ Methods that extend this generic function should
       (if (cl-loop
            with nlines = 0
            for c across res when (eq c ?\n) do (cl-incf nlines)
-           thereis (> nlines 10))
+           thereis (> nlines jupyter-display-short-eval-result-max-lines))
           (jupyter-with-display-buffer "result" 'reset
             (insert res)
             (goto-char (point-min))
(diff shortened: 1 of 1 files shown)


dzop:
Thanks!

Some comments:

1. `jupyter-display-output-function` is too general a name since the changes only apply to displaying the output of short evaluation results. Maybe something like `jupyter-display-short-eval-result-function`.

2. It would be simpler to change the variable type to a function instead of a symbol so that we don't add in the extra `popup` dependency. Then you would simply do
`(funcall jupyter-display-...-function s)`.
    - In this case you should use function quoting for the default value of the variable, change `'message` into `#'message`.

3. I don't think we need to add in the `jupyter--display-output` function, I think it would be better to just do the work in `jupyter--display-eval-result`. If you make the changes in (2) then you can just do
`(funcall jupyter-display-...-function (format "%s" res))`

sebastianpech:
Good idea to get rid fo the dependency. I implemented your comments.

How about making the limit of 10 lines also changeable by the user? In the minibuffer more than 10 lines is arguably a good limit, but for tooltips I would prefer more lines.

dzop:
Sounds good. What about changing the variable name to `jupyter-display-short-eval-result-max-lines` which makes its purpose slightly more obvious and a little more conventional since there are variables such as `org-babel-min-lines-for-block-output` and `messages-buffer-max-lines`.

About the documentation strings, can you make sure that the first line is a complete sentence that explains what each variable does as succinct as possible. For example, for the max lines variables you can have the first line in the documentation string say

```Maximum number of lines for eval results to still be considered short.```

and then use the rest of the documentation string to explain what short means.

Also, could you squash the first two commits and force push to get rid of the commit that has the `popup` dependency.

**Comment by sebastianpech** [comment]
I'll do that. Do you have any idea why the second commit failed?

**Comment by sebastianpech** [comment]
Ok, renamed, rephrased and squashed.

**Comment by dzop** [comment]
Some more changes:

- Could you ensure that the maximum line length is 80 characters in the documentation strings. This is not a hard limit, but 90 is certainly too long considering the rest of the code.

- In the documentation of `jupyter-display-short-eval-result-function`
    - `cosidered` -> `considered`
    - `in case` -> `when`
    - there is no need for the redundant `lines` after `jupyter-display-short-eval-result-max-lines`
        - you should also remove the extra `lines` in `jupyter-eval-string`
    - `` `popup'`` -> ``the `popup' package``
    - the last sentence should end in a period

Also, I'm thinking it would be better to do the following renamings:

- `jupyter-display-short-eval-result-funtion` -> `jupyter-eval-short-result-display-function`
- `jupyter-display-short-eval-result-max-lines` -> `jupyter-eval-short-result-max-lines`

Since these variables are associated with evaluation, it makes more sense to give them the `jupyter-eval` prefix as opposed to `jupyter-display` which is not really used anywhere. What do you think?

**Comment by sebastianpech** [comment]
That seems reasonable. I renamed the variables

File: jupyter-client.el [hunk 1]
```diff
@@ -37,22 +37,23 @@
 (require 'jupyter-mime)
 (require 'jupyter-messages)
 
-(defcustom jupyter-display-short-eval-result-function #'message
```
dzop: I would prefer the name to be `jupyter-eval-short-result-display-function`. I don't think there is enough context in `jupyter-short-result-function` to understand what it is doing.

sebastianpech: That's what I changed it to.

dzop: Duh. Sorry I was confused by the diff.

PR #45, opened 2019-02-28
Port scimax's source block helper functions

UndeadKernel:
This is a work in progress. Things remaining:
1) Is the location of the function definition adequate? How about the naming convention I used?
1) Cannot figure out how to ask the kernel to restart without getting an error.
1) Do we want default bindings for some of the most useful functions added to `jupyter-org-interaction-mode-map`?
1) Is the key binding of the hydra bound to the right key map? (current it's bound to the local map, i.e., org-mode). The Hydra should be callable also outside source blocks (so `jupyter-org-interaction-mode-map` is not a good candidate).
1) Some functions depend on `avy`, i.e., `jupyter-org-jump-to-visible-block`. How do we deal with this? Easiest could be to just remove the functions if they are not suitable to `emacs-jupyter`. They are useful but we could live without them.
1) Code style can be improved (once all other problems are defined I can work on cleaning up).

We provide a hydra to easily use all the new functions.
By default, the binding to call the hydra is `C-c h`
The hydra is only created if the hydra package is loaded.


jupyter-org-client.el
```diff
@@ -416,6 +416,357 @@ the `syntax-table' will be set to that of the REPL buffers."
...
+                    ""))
+            (list next-src-block current-src-block))
+      ;; Now create the merged block, point is where the current block was
+      (insert (format "#+BEGIN_SRC %s %s %s\n%s#+END_SRC\n\n"
+                      lang switches parameters merged-code))
+      (forward-line -3)
+      (end-of-line))))
+
+;;;###autoload
+(defun jupyter-org-move-src-block (&optional below)
```
dzop: Can `org-element-swap-A-B` used to do the work?

jupyter-org-client.el
```diff
@@ -416,6 +416,357 @@ the `syntax-table' will be set to that of the REPL buffers."
...
+  (save-excursion
+    (goto-char (point-min))
+    (while (org-babel-next-src-block)
+      (org-babel-remove-result))))
+
+;;;###autoload
+(defun jupyter-org-restart-kernel-execute-block ()
+  "Restart kernel and execute block."
+  (interactive)
+  (let* ((params (car jupyter-org--src-block-cache))
```
dzop: This looks like a job for `jupyter-org-with-src-block-client`.

jupyter-org-client.el
```diff
@@ -416,6 +416,357 @@ the `syntax-table' will be set to that of the REPL buffers."
...
+         (jupyter-current-client
+          (buffer-local-value 'jupyter-current-client
+                              (org-babel-jupyter-initiate-session
+                               (alist-get :session params) params))))
+    ;; FIXME: retarting the kernel does not work
+    (jupyter-repl-restart-kernel))
```
dzop: What is the error that you are getting?

jupyter-org-client.el
```diff
@@ -416,6 +416,357 @@ the `syntax-table' will be set to that of the REPL buffers."
...
+      ;; Remove source blocks
+      (mapc (lambda (src-block)
+              (goto-char (org-element-property :begin src-block))
+              (org-babel-remove-result)
+              (setf (buffer-substring (org-element-property :begin src-block)
+                                      (org-element-property :end src-block))
+                    ""))
+            (list next-src-block current-src-block))
+      ;; Now create the merged block, point is where the current block was
+      (insert (format "#+BEGIN_SRC %s %s %s\n%s#+END_SRC\n\n"
```
dzop: I would prefer this to be

```emacs-lisp
(insert
 (org-element-interpret-data
  (org-element-put-property
   (jupyter-org-src-block lang parameters merged-code switches)
   :post-blank 1)))
```

The way you have it now is fine, it is just that I already create `org` syntax trees and convert to a string using `org-element-interpret-data` everywhere else.

jupyter-org-client.el
```diff
@@ -416,6 +416,357 @@ the `syntax-table' will be set to that of the REPL buffers."
...
+
+;;;###autoload
+(defun jupyter-org-clone-block (&optional below)
+  "Clone the block."
+  (interactive "P")
+  (let* ((src (org-element-context))
+         (code (org-element-property :value src)))
+    (unless (eq (org-element-type src) 'src-block)
+      (error "Not in a source block"))
+    (jupyter-org-insert-src-block (not below))
```
dzop: We can just use `(save-excursion (insert (org-element-interpret-data src)))` here right? After moving to the right location to insert the source block. This would remove the need for the rest of the lines below.

jupyter-org-client.el
```diff
@@ -416,6 +416,357 @@ the `syntax-table' will be set to that of the REPL buffers."
...
+             location)
+        (if below
+            (progn
+              (goto-char start)
+              (setq location (org-babel-where-is-src-block-result nil nil))
+              (if (not  location)
+                  (goto-char end)
+                (goto-char location)
+                (goto-char (org-element-property :end (org-element-context))))
+              (insert (format "#+BEGIN_SRC %s\n\n#+END_SRC\n\n"
```
dzop: You can let `org-element-interpret-data` do most of the work here.

```emacs-lisp
(insert
 (org-element-interpret-data
  (org-element-put-property src :value "")))
```

should make all of the lang, switches, and parameters variables unnecessary.

jupyter-org-client.el
```diff
@@ -416,6 +416,357 @@ the `syntax-table' will be set to that of the REPL buffers."
...
+                              (concat lang (and lang " ")
+                                      switches (and switches " ")
+                                      parameters (and parameters " "))))
+              (forward-line -3))
+          ;; after current block
+          (goto-char (org-element-property :begin (org-element-context)))
+          (insert (format "#+BEGIN_SRC %s\n\n#+END_SRC\n\n"
```
dzop: See above comment.

jupyter-org-client.el
```diff
@@ -416,6 +416,357 @@ the `syntax-table' will be set to that of the REPL buffers."
...
+                          (concat lang (and lang " ")
+                                  switches (and switches " ")
+                                  parameters (and parameters " "))))
+          (forward-line -3)))
+
+    ;; Not in a src block, just insert a block
+    (beginning-of-line)
+    (insert (format "#+BEGIN_SRC %s\n\n#+END_SRC\n" (completing-read "Language: " (mapcar 'car org-babel-load-languages))))
```
dzop: Again, I would like to use `org-element-interpret-data`, see https://github.com/dzop/emacs-jupyter/pull/45/files#r261337038

jupyter-org-client.el
```diff
@@ -416,6 +416,357 @@ the `syntax-table' will be set to that of the REPL buffers."
...
+    (forward-line -1)))
+
+;;;###autoload
+(defun jupyter-org-split-src-block (&optional below)
```
dzop: Can `org-babel-demarcate-block` be used to do most of the work here?

dzop:
> Is the location of the function definition adequate?

The `jupyter-org-client.el` file is meant to hold the code necessary to provide an `org-mode` Jupyter client whereas `ob-jupyter.el` is used to hold the code necessary to interface with `org-babel`. I'm thinking these functions should be put into another file since they are extensions that give similar behavior to Jupyter notebooks. Maybe a file like `jupyter-org-x.el` to indicate that these extra functions are for interacting with Jupyter source blocks.

> Do we want default bindings for some of the most useful functions added to `jupyter-org-interaction-mode-map`?

That makes sense. It depends on what people find to be the most useful functions though. So maybe hold off until there is more feedback.

> Is the key binding of the hydra bound to the right key map? (current it's bound to the local map, i.e., org-mode). The Hydra should be callable also outside source blocks (so `jupyter-org-interaction-mode-map` is not a good candidate).

The `jupyter-org-interaction-mode-map` can be used as long as you just use `define-key` to bind it. If `jupyter-org-define-key` is used, then the Hydra will only be callable inside a source block. Through the regular `define-key` it will be callable inside or outside a code block as long as `jupyter-org-interaction-mode` is enabled. 

> Some functions depend on `avy`, i.e., `jupyter-org-jump-to-visible-block`. How do we deal with this? Easiest could be to just remove the functions if they are not suitable to `emacs-jupyter`. They are useful but we could live without them.

I'd have to think about this. I would rather not depend on extra packages since the goal of `emacs-jupyter` is to integrate Jupyter with Emacs' built-in features. My strategy for supporting external packages so far has been to only enable support for a feature that depends on an external package when that package is already available.

jupyter-org-client.el
```diff
@@ -416,6 +416,357 @@ the `syntax-table' will be set to that of the REPL buffers."
...
+  (let* ((params (car jupyter-org--src-block-cache))
+         (jupyter-current-client
+          (buffer-local-value 'jupyter-current-client
+                              (org-babel-jupyter-initiate-session
+                               (alist-get :session params) params))))
+    ;; FIXME: retarting the kernel does not work
+    (jupyter-repl-restart-kernel))
+  (org-babel-execute-src-block-maybe))
+
+(with-eval-after-load 'hydra
```
dzop: I would prefer not to use `with-eval-after-load` since that is mainly meant for init files and not necessarily libraries.

I don't really know if there is a clean way to do this without `with-eval-after-load` so I'm thinking it may be best to make a new package that explicitly depends on `hydra` and then we point the user to load that package in their init files? This wouldn't be very convenient though.

dzop: Maybe we can do something like

```emacs-lisp
(defun jupyter-org-hydra/body ()
  (interactive)
  (unless (require 'hydra nil t)
    (error "Install the `hydra' package."))
  ;; Unbind this function now since it will be re-bound to the actual hydra
  ;; below
  (fmakunbound 'jupyter-org-hydra/body)
  (eval `(defhydra ...)))
```

jupyter-org-client.el
UndeadKernel: The problem is that we need to also move the results of a block (if any).
Using `org-element-swap-A-B` would rather complicate things.

jupyter-org-client.el
UndeadKernel: With the following code all seems to work ok
```lisp
(jupyter-org-with-src-block-client
   (jupyter-repl-restart-kernel))
```
However, no matter how I restart the kernel, I get some messages in the repl's stderr. The buffer name of the stderr is "*jupyter-*jupyter-repl[python 3.7.2]-testpy2-stderr**" and contains:
```
ERROR:root:Invalid alias: The name clear can't be aliased because it is another
magic command.
ERROR:root:Invalid alias: The name more can't be aliased because it is another
magic command.
ERROR:root:Invalid alias: The name less can't be aliased because it is another
magic command.
ERROR:root:Invalid alias: The name man can't be aliased because it is another
magic command.
```
And the following error message:
```log
Restarting kernel...
Error running timer ‘jupyter-handle-message’: (error "Found end of previous cell")
python3 kernel shutting down...done
Starting python3 kernel process...done
```

jupyter-org-client.el
UndeadKernel: See my comment above.
Even using the "official" way of typing `C-c C-r` inside a source block yields the same errors.

jupyter-org-client.el
UndeadKernel: This works great. `jupyter-org-src-block` reduces the work I needed to do to make sure the resulting source block was correctly structured.

jupyter-org-client.el
UndeadKernel: I cannot quite use `org-element-interpret-data` or, rather, I prefer using `jupyter-org-insert-src-block` as then I don't have to go through the logic of figuring out the position where to clone a block. Finding the position is tricky because I need to check for results.

UndeadKernel: I however saw that the logic to move to the next block was wrong and is not needed.

jupyter-org-client.el
UndeadKernel: After switching to `org-element-interpret-data`, I'm left with a sour feeling knowing that "src_block" is inserted all in lower case letters instead of the pretty "SRC_BLOCK".
Is this something we should change? I prefer the uppercase form.

jupyter-org-client.el
UndeadKernel: I've switched to `org-element-interpret-data`. 
The only remaining questions is if we do anything with the "src_block" vs "SRC_BLOCK".
In my version, I was using uppercase letters. The `org-element-interpret-data` function uses only lower case letters.

jupyter-org-client.el
UndeadKernel: Indeed. I have switched the logic to use `org-babel-demarcate-block`.

jupyter-org-client.el
UndeadKernel: Quite a nice idea.
I'll implement that.

jupyter-org-extensions.el
```diff
@@ -0,0 +1,404 @@
...
+           (org-element-interpret-data
+            (org-element-put-property
+             (jupyter-org-src-block lang parameters "\n" switches)
+             :post-blank 1)))
+          (forward-line -3)))
+
+    ;; not in a src block, insert a new block, query for jupyter kernel
+    (beginning-of-line)
+    (let ((kernel
+           (completing-read "Kernel: "
```
dzop: There is also the function `jupyter-completing-read-kernelspec` which does the same thing.

jupyter-org-extensions.el
```diff
@@ -0,0 +1,404 @@
...
+(defun jupyter-org-move-src-block (&optional below)
+  "Move source block before of after another.
+
+If BELOW is non-nil, move the block down, otherwise move it up."
+  (interactive)
+  ;;skip all this if there are no previous or next source block
+  (when (condition-case nil
+            (save-excursion
+              (if below
+                  (org-babel-next-src-block)
```
dzop: What do we want to do about subtree boundaries? `org-babel-{next|previous}-src-block` do not consider subtree boundaries, but does it make sense to move a code block across those boundaries?

For example if we have

```
* A

** B

#+BEGIN_SRC jupyter-python
1 + 1
#+END_SRC

* C

#+BEGIN_SRC jupyter-python
1 + 2
#+END_SRC
```

it is currently possible to move the second source block before the first, but should this be possible? In `org-mode` you can only move subtrees around, e.g. using `org-move-subtree-down`, that are on the same level. I think it would be more consistent with how `org-mode` works to prevent moving source blocks across different subtree levels. Granted source blocks do not have a level like subtrees do.

jupyter-org-extensions.el
```diff
@@ -0,0 +1,404 @@
...
+      (end-of-line))))
+
+;;;###autoload
+(defun jupyter-org-execute-and-next-block (&optional new)
+  "Execute this block and jump or add a new one.
+
+If a new block is created, use the same language, switches and parameters.
+With prefix arg NEW, always insert new cell."
+  (interactive "P")
+  (unless (eq (org-element-type (org-element-context)) 'src-block)
```
dzop: I think it would be better to use `org-in-src-block-p` here.

jupyter-org-extensions.el
```diff
@@ -0,0 +1,404 @@
...
+    (forward-line -2)))
+
+;;;###autoload
+(defun jupyter-org-split-src-block (&optional below)
+  "Split the current src block with point in upper block.
+
+With a prefix BELOW move point to lower block."
+  (interactive "P")
+  (let ((src-block (org-element-context)))
+    (unless (eq (org-element-type src-block) 'src-block)
```
dzop: I think it would be better to use `org-in-src-block-p` here.

jupyter-org-extensions.el
```diff
@@ -0,0 +1,404 @@
...
+        (save-excursion
+          (goto-char result-start)
+          (setq end (org-element-property :end src)))
+      (setq end (org-element-property :end src)))
+    (kill-region
+     (org-element-property :begin src)
+     end)))
+
+;;;###autoload
+(defun jupyter-org-copy-block-and-results ()
```
dzop: What's the difference between this function and `jupyter-org-kill-block-and-results`?

jupyter-org-client.el
dzop: I don't know about all of the Python errors, seems like more of a Python issue. I am on Python 3.7.1 and I haven't been getting those errors when I testing the kernel restart stuff. Can you try 1f5c7743d534929efaad7592762f241700bc1f6c to see if that restarts the kernel more cleanly.

jupyter-org-client.el
dzop: That is unfortunate, the problem with hard-coding uppercase here is what if someone has all their code blocks in all lowercase which seems to be the default behavior now in `org-mode` 9.2 (https://lists.gnu.org/archive/html/emacs-orgmode/2019-02/msg00148.html).

I guess the simplest solution would be to avoid using `org-element-interpret-data` and have a customizable variable, `jupyter-org-prefer-uppercase-src-blocks` which defaults to 
`(version< (org-version) "9.2")` and use that to determine if uppercase or lowercase block delimiters should be inserted.

jupyter-org-extensions.el
```diff
@@ -0,0 +1,404 @@
...
+      (error "Not in a source block"))
+    (jupyter-org-insert-src-block below)
+    (delete-char 1)
+    (insert code)
+    ;; move to the end of the last line of the cloned block
+    (forward-line -1)
+    (end-of-line)))
+
+;;;###autoload
+(defun jupyter-org-merge-blocks ()
```
dzop: This can be simplified by just working with the text in the buffer instead of pulling out all of the `:language`, `:switches`, etc. from the element. For example, we can delete and save the text in between the two source blocks using `delete-and-extract-region` on the region 

`((org-element-property :end current-src-block) . (org-element-property :begin next-src-block))`

then delete the `#+END_SRC ... #+BEGIN_SRC` and the two source blocks will be merged. Of course this is after deleting any results the source blocks might have. Then we move to the end of the merged source block and insert the saved text that was between the two.

Doing this also avoids having to consider if the source block delimiters are uppercase or lowercase as well.

WDYT?

jupyter-org-extensions.el
```diff
@@ -0,0 +1,404 @@
...
+              (save-excursion
+                (goto-char results-start)
+                (goto-char (org-babel-result-end))
+                ;; if line is empty, take that empty line by moving down
+                (when (looking-at-p "[[:space:]]*$")
+                  (forward-line 1))
+                (point)))))
+      ;; kill from the start of the source block to the end of the results
+      ;; ... if there are no results, to the end of the source block
+      (kill-region
```
dzop: I think it would be good to use `delete-and-extract-region` here. Is there a reason to use `kill-region`?

jupyter-org-client.el
UndeadKernel: With the latest master, the kernel restarts more cleanly (less error messages) but there are still some error messages:
```
Error running timer ‘jupyter-handle-message’: (error "Cell for request not found") [4 times]
```
The trace I get for the error is this one:
```lisp
Debugger entered--Lisp error: (error "Cell for request not found")
  signal(error ("Cell for request not found"))
  error("Cell for request not found")
  jupyter-repl-goto-cell(#s(jupyter-request :id "cfd3b78d-1ec6-409b-ba97-41c2cb82c655" :time (23677 1564 171792 52000) :idle-received-p t :last-message (:header (message-part "{\"msg_id\":\"e8c6a08d-83b1c9f7d74a3dfb1b80248e\",\"msg_type\":\"stream\",\"username\":\"boy\",\"session\":\"464ef51b-d62e8feaf3cf3b8548035101\",\"date\":\"2019-03-04T11:03:56.326894Z\",\"version\":\"5.3\"}" (:msg_id "e8c6a08d-83b1c9f7d74a3dfb1b80248e" :msg_type :stream :username "boy" :session "464ef51b-d62e8feaf3cf3b8548035101" :date "2019-03-04T11:03:56.326894Z" :version "5.3")) :msg_id "e8c6a08d-83b1c9f7d74a3dfb1b80248e" :msg_type :stream :parent_header (message-part "{\"msg_id\":\"cfd3b78d-1ec6-409b-ba97-41c2cb82c655\",\"msg_type\":\"shutdown_request\",\"version\":\"5.3\",\"username\":\"boy\",\"session\":\"9b3dd7de-5a5f-4878-8643-a179c8a4bee2\",\"date\":\"2019-03-04T12:03:56.171874+01:00\"}" (:msg_id "cfd3b78d-1ec6-409b-ba97-41c2cb82c655" :msg_type :shutdown-request :version "5.3" :username "boy" :session "9b3dd7de-5a5f-4878-8643-a179c8a4bee2" :date "2019-03-04T12:03:56.171874+01:00")) :metadata (message-part "{}" nil) :content (message-part "{\"name\":\"stderr\",\"text\":\"ERROR:root:Invalid alias: The name clear can't be aliased because it is another magic command.\\n\"}" (:name "stderr" :text "ERROR:root:Invalid alias: The name clear can't be aliased because it is another magic command.\n")) :buffers nil) :inhibited-handlers nil :callbacks ((:shutdown-reply . #f(compiled-function (msg) #<bytecode 0x1240cc1>)))))
  #f(compiled-function (client req name text) #<bytecode 0x23c09d5>)(#<jupyter-org-client jupyter-org-client> #s(jupyter-request :id "cfd3b78d-1ec6-409b-ba97-41c2cb82c655" :time (23677 1564 171792 52000) :idle-received-p t :last-message (:header (message-part "{\"msg_id\":\"e8c6a08d-83b1c9f7d74a3dfb1b80248e\",\"msg_type\":\"stream\",\"username\":\"boy\",\"session\":\"464ef51b-d62e8feaf3cf3b8548035101\",\"date\":\"2019-03-04T11:03:56.326894Z\",\"version\":\"5.3\"}" (:msg_id "e8c6a08d-83b1c9f7d74a3dfb1b80248e" :msg_type :stream :username "boy" :session "464ef51b-d62e8feaf3cf3b8548035101" :date "2019-03-04T11:03:56.326894Z" :version "5.3")) :msg_id "e8c6a08d-83b1c9f7d74a3dfb1b80248e" :msg_type :stream :parent_header (message-part "{\"msg_id\":\"cfd3b78d-1ec6-409b-ba97-41c2cb82c655\",\"msg_type\":\"shutdown_request\",\"version\":\"5.3\",\"username\":\"boy\",\"session\":\"9b3dd7de-5a5f-4878-8643-a179c8a4bee2\",\"date\":\"2019-03-04T12:03:56.171874+01:00\"}" (:msg_id "cfd3b78d-1ec6-409b-ba97-41c2cb82c655" :msg_type :shutdown-request :version "5.3" :username "boy" :session "9b3dd7de-5a5f-4878-8643-a179c8a4bee2" :date "2019-03-04T12:03:56.171874+01:00")) :metadata (message-part "{}" nil) :content (message-part "{\"name\":\"stderr\",\"text\":\"ERROR:root:Invalid alias: The name clear can't be aliased because it is another magic command.\\n\"}" (:name "stderr" :text "ERROR:root:Invalid alias: The name clear can't be aliased because it is another magic command.\n")) :buffers nil) :inhibited-handlers nil :callbacks ((:shutdown-reply . #f(compiled-function (msg) #<bytecode 0x1240cc1>)))) "stderr" "ERROR:root:Invalid alias: The name clear can't be aliased because it is another magic command.\n")
  apply(#f(compiled-function (client req name text) #<bytecode 0x23c09d5>) #<jupyter-org-client jupyter-org-client> (#s(jupyter-request :id "cfd3b78d-1ec6-409b-ba97-41c2cb82c655" :time (23677 1564 171792 52000) :idle-received-p t :last-message (:header (message-part "{\"msg_id\":\"e8c6a08d-83b1c9f7d74a3dfb1b80248e\",\"msg_type\":\"stream\",\"username\":\"boy\",\"session\":\"464ef51b-d62e8feaf3cf3b8548035101\",\"date\":\"2019-03-04T11:03:56.326894Z\",\"version\":\"5.3\"}" (:msg_id "e8c6a08d-83b1c9f7d74a3dfb1b80248e" :msg_type :stream :username "boy" :session "464ef51b-d62e8feaf3cf3b8548035101" :date "2019-03-04T11:03:56.326894Z" :version "5.3")) :msg_id "e8c6a08d-83b1c9f7d74a3dfb1b80248e" :msg_type :stream :parent_header (message-part "{\"msg_id\":\"cfd3b78d-1ec6-409b-ba97-41c2cb82c655\",\"msg_type\":\"shutdown_request\",\"version\":\"5.3\",\"username\":\"boy\",\"session\":\"9b3dd7de-5a5f-4878-8643-a179c8a4bee2\",\"date\":\"2019-03-04T12:03:56.171874+01:00\"}" (:msg_id "cfd3b78d-1ec6-409b-ba97-41c2cb82c655" :msg_type :shutdown-request :version "5.3" :username "boy" :session "9b3dd7de-5a5f-4878-8643-a179c8a4bee2" :date "2019-03-04T12:03:56.171874+01:00")) :metadata (message-part "{}" nil) :content (message-part "{\"name\":\"stderr\",\"text\":\"ERROR:root:Invalid alias: The name clear can't be aliased because it is another magic command.\\n\"}" (:name "stderr" :text "ERROR:root:Invalid alias: The name clear can't be aliased because it is another magic command.\n")) :buffers nil) :inhibited-handlers nil :callbacks ((:shutdown-reply . #f(compiled-function (msg) #<bytecode 0x1240cc1>)))) "stderr" "ERROR:root:Invalid alias: The name clear can't be aliased because it is another magic command.\n"))
  #f(compiled-function (arg &rest args) #<bytecode 0x12d69c5>)(#<jupyter-org-client jupyter-org-client> #s(jupyter-request :id "cfd3b78d-1ec6-409b-ba97-41c2cb82c655" :time (23677 1564 171792 52000) :idle-received-p t :last-message (:header (message-part "{\"msg_id\":\"e8c6a08d-83b1c9f7d74a3dfb1b80248e\",\"msg_type\":\"stream\",\"username\":\"boy\",\"session\":\"464ef51b-d62e8feaf3cf3b8548035101\",\"date\":\"2019-03-04T11:03:56.326894Z\",\"version\":\"5.3\"}" (:msg_id "e8c6a08d-83b1c9f7d74a3dfb1b80248e" :msg_type :stream :username "boy" :session "464ef51b-d62e8feaf3cf3b8548035101" :date "2019-03-04T11:03:56.326894Z" :version "5.3")) :msg_id "e8c6a08d-83b1c9f7d74a3dfb1b80248e" :msg_type :stream :parent_header (message-part "{\"msg_id\":\"cfd3b78d-1ec6-409b-ba97-41c2cb82c655\",\"msg_type\":\"shutdown_request\",\"version\":\"5.3\",\"username\":\"boy\",\"session\":\"9b3dd7de-5a5f-4878-8643-a179c8a4bee2\",\"date\":\"2019-03-04T12:03:56.171874+01:00\"}" (:msg_id "cfd3b78d-1ec6-409b-ba97-41c2cb82c655" :msg_type :shutdown-request :version "5.3" :username "boy" :session "9b3dd7de-5a5f-4878-8643-a179c8a4bee2" :date "2019-03-04T12:03:56.171874+01:00")) :metadata (message-part "{}" nil) :content (message-part "{\"name\":\"stderr\",\"text\":\"ERROR:root:Invalid alias: The name clear can't be aliased because it is another magic command.\\n\"}" (:name "stderr" :text "ERROR:root:Invalid alias: The name clear can't be aliased because it is another magic command.\n")) :buffers nil) :inhibited-handlers nil :callbacks ((:shutdown-reply . #f(compiled-function (msg) #<bytecode 0x1240cc1>)))) "stderr" "ERROR:root:Invalid alias: The name clear can't be aliased because it is another magic command.\n")
  apply(#f(compiled-function (arg &rest args) #<bytecode 0x12d69c5>) #<jupyter-org-client jupyter-org-client> #s(jupyter-request :id "cfd3b78d-1ec6-409b-ba97-41c2cb82c655" :time (23677 1564 171792 52000) :idle-received-p t :last-message (:header (message-part "{\"msg_id\":\"e8c6a08d-83b1c9f7d74a3dfb1b80248e\",\"msg_type\":\"stream\",\"username\":\"boy\",\"session\":\"464ef51b-d62e8feaf3cf3b8548035101\",\"date\":\"2019-03-04T11:03:56.326894Z\",\"version\":\"5.3\"}" (:msg_id "e8c6a08d-83b1c9f7d74a3dfb1b80248e" :msg_type :stream :username "boy" :session "464ef51b-d62e8feaf3cf3b8548035101" :date "2019-03-04T11:03:56.326894Z" :version "5.3")) :msg_id "e8c6a08d-83b1c9f7d74a3dfb1b80248e" :msg_type :stream :parent_header (message-part "{\"msg_id\":\"cfd3b78d-1ec6-409b-ba97-41c2cb82c655\",\"msg_type\":\"shutdown_request\",\"version\":\"5.3\",\"username\":\"boy\",\"session\":\"9b3dd7de-5a5f-4878-8643-a179c8a4bee2\",\"date\":\"2019-03-04T12:03:56.171874+01:00\"}" (:msg_id "cfd3b78d-1ec6-409b-ba97-41c2cb82c655" :msg_type :shutdown-request :version "5.3" :username "boy" :session "9b3dd7de-5a5f-4878-8643-a179c8a4bee2" :date "2019-03-04T12:03:56.171874+01:00")) :metadata (message-part "{}" nil) :content (message-part "{\"name\":\"stderr\",\"text\":\"ERROR:root:Invalid alias: The name clear can't be aliased because it is another magic command.\\n\"}" (:name "stderr" :text "ERROR:root:Invalid alias: The name clear can't be aliased because it is another magic command.\n")) :buffers nil) :inhibited-handlers nil :callbacks ((:shutdown-reply . #f(compiled-function (msg) #<bytecode 0x1240cc1>)))) ("stderr" "ERROR:root:Invalid alias: The name clear can't be aliased because it is another magic command.\n"))
  jupyter-handle-stream(#<jupyter-org-client jupyter-org-client> #s(jupyter-request :id "cfd3b78d-1ec6-409b-ba97-41c2cb82c655" :time (23677 1564 171792 52000) :idle-received-p t :last-message (:header (message-part "{\"msg_id\":\"e8c6a08d-83b1c9f7d74a3dfb1b80248e\",\"msg_type\":\"stream\",\"username\":\"boy\",\"session\":\"464ef51b-d62e8feaf3cf3b8548035101\",\"date\":\"2019-03-04T11:03:56.326894Z\",\"version\":\"5.3\"}" (:msg_id "e8c6a08d-83b1c9f7d74a3dfb1b80248e" :msg_type :stream :username "boy" :session "464ef51b-d62e8feaf3cf3b8548035101" :date "2019-03-04T11:03:56.326894Z" :version "5.3")) :msg_id "e8c6a08d-83b1c9f7d74a3dfb1b80248e" :msg_type :stream :parent_header (message-part "{\"msg_id\":\"cfd3b78d-1ec6-409b-ba97-41c2cb82c655\",\"msg_type\":\"shutdown_request\",\"version\":\"5.3\",\"username\":\"boy\",\"session\":\"9b3dd7de-5a5f-4878-8643-a179c8a4bee2\",\"date\":\"2019-03-04T12:03:56.171874+01:00\"}" (:msg_id "cfd3b78d-1ec6-409b-ba97-41c2cb82c655" :msg_type :shutdown-request :version "5.3" :username "boy" :session "9b3dd7de-5a5f-4878-8643-a179c8a4bee2" :date "2019-03-04T12:03:56.171874+01:00")) :metadata (message-part "{}" nil) :content (message-part "{\"name\":\"stderr\",\"text\":\"ERROR:root:Invalid alias: The name clear can't be aliased because it is another magic command.\\n\"}" (:name "stderr" :text "ERROR:root:Invalid alias: The name clear can't be aliased because it is another magic command.\n")) :buffers nil) :inhibited-handlers nil :callbacks ((:shutdown-reply . #f(compiled-function (msg) #<bytecode 0x1240cc1>)))) "stderr" "ERROR:root:Invalid alias: The name clear can't be aliased because it is another magic command.\n")
  #f(compiled-function (channel client req msg) #<bytecode 0x232c895>)(:iopub #<jupyter-org-client jupyter-org-client> #s(jupyter-request :id "cfd3b78d-1ec6-409b-ba97-41c2cb82c655" :time (23677 1564 171792 52000) :idle-received-p t :last-message (:header (message-part "{\"msg_id\":\"e8c6a08d-83b1c9f7d74a3dfb1b80248e\",\"msg_type\":\"stream\",\"username\":\"boy\",\"session\":\"464ef51b-d62e8feaf3cf3b8548035101\",\"date\":\"2019-03-04T11:03:56.326894Z\",\"version\":\"5.3\"}" (:msg_id "e8c6a08d-83b1c9f7d74a3dfb1b80248e" :msg_type :stream :username "boy" :session "464ef51b-d62e8feaf3cf3b8548035101" :date "2019-03-04T11:03:56.326894Z" :version "5.3")) :msg_id "e8c6a08d-83b1c9f7d74a3dfb1b80248e" :msg_type :stream :parent_header (message-part "{\"msg_id\":\"cfd3b78d-1ec6-409b-ba97-41c2cb82c655\",\"msg_type\":\"shutdown_request\",\"version\":\"5.3\",\"username\":\"boy\",\"session\":\"9b3dd7de-5a5f-4878-8643-a179c8a4bee2\",\"date\":\"2019-03-04T12:03:56.171874+01:00\"}" (:msg_id "cfd3b78d-1ec6-409b-ba97-41c2cb82c655" :msg_type :shutdown-request :version "5.3" :username "boy" :session "9b3dd7de-5a5f-4878-8643-a179c8a4bee2" :date "2019-03-04T12:03:56.171874+01:00")) :metadata (message-part "{}" nil) :content (message-part "{\"name\":\"stderr\",\"text\":\"ERROR:root:Invalid alias: The name clear can't be aliased because it is another magic command.\\n\"}" (:name "stderr" :text "ERROR:root:Invalid alias: The name clear can't be aliased because it is another magic command.\n")) :buffers nil) :inhibited-handlers nil :callbacks ((:shutdown-reply . #f(compiled-function (msg) #<bytecode 0x1240cc1>)))) (:header (message-part "{\"msg_id\":\"e8c6a08d-83b1c9f7d74a3dfb1b80248e\",\"msg_type\":\"stream\",\"username\":\"boy\",\"session\":\"464ef51b-d62e8feaf3cf3b8548035101\",\"date\":\"2019-03-04T11:03:56.326894Z\",\"version\":\"5.3\"}" (:msg_id "e8c6a08d-83b1c9f7d74a3dfb1b80248e" :msg_type :stream :username "boy" :session "464ef51b-d62e8feaf3cf3b8548035101" :date "2019-03-04T11:03:56.326894Z" :version "5.3")) :msg_id "e8c6a08d-83b1c9f7d74a3dfb1b80248e" :msg_type :stream :parent_header (message-part "{\"msg_id\":\"cfd3b78d-1ec6-409b-ba97-41c2cb82c655\",\"msg_type\":\"shutdown_request\",\"version\":\"5.3\",\"username\":\"boy\",\"session\":\"9b3dd7de-5a5f-4878-8643-a179c8a4bee2\",\"date\":\"2019-03-04T12:03:56.171874+01:00\"}" (:msg_id "cfd3b78d-1ec6-409b-ba97-41c2cb82c655" :msg_type :shutdown-request :version "5.3" :username "boy" :session "9b3dd7de-5a5f-4878-8643-a179c8a4bee2" :date "2019-03-04T12:03:56.171874+01:00")) :metadata (message-part "{}" nil) :content (message-part "{\"name\":\"stderr\",\"text\":\"ERROR:root:Invalid alias: The name clear can't be aliased because it is another magic command.\\n\"}" (:name "stderr" :text "ERROR:root:Invalid alias: The name clear can't be aliased because it is another magic command.\n")) :buffers nil))
  apply(#f(compiled-function (channel client req msg) #<bytecode 0x232c895>) :iopub (#<jupyter-org-client jupyter-org-client> #s(jupyter-request :id "cfd3b78d-1ec6-409b-ba97-41c2cb82c655" :time (23677 1564 171792 52000) :idle-received-p t :last-message (:header (message-part "{\"msg_id\":\"e8c6a08d-83b1c9f7d74a3dfb1b80248e\",\"msg_type\":\"stream\",\"username\":\"boy\",\"session\":\"464ef51b-d62e8feaf3cf3b8548035101\",\"date\":\"2019-03-04T11:03:56.326894Z\",\"version\":\"5.3\"}" (:msg_id "e8c6a08d-83b1c9f7d74a3dfb1b80248e" :msg_type :stream :username "boy" :session "464ef51b-d62e8feaf3cf3b8548035101" :date "2019-03-04T11:03:56.326894Z" :version "5.3")) :msg_id "e8c6a08d-83b1c9f7d74a3dfb1b80248e" :msg_type :stream :parent_header (message-part "{\"msg_id\":\"cfd3b78d-1ec6-409b-ba97-41c2cb82c655\",\"msg_type\":\"shutdown_request\",\"version\":\"5.3\",\"username\":\"boy\",\"session\":\"9b3dd7de-5a5f-4878-8643-a179c8a4bee2\",\"date\":\"2019-03-04T12:03:56.171874+01:00\"}" (:msg_id "cfd3b78d-1ec6-409b-ba97-41c2cb82c655" :msg_type :shutdown-request :version "5.3" :username "boy" :session "9b3dd7de-5a5f-4878-8643-a179c8a4bee2" :date "2019-03-04T12:03:56.171874+01:00")) :metadata (message-part "{}" nil) :content (message-part "{\"name\":\"stderr\",\"text\":\"ERROR:root:Invalid alias: The name clear can't be aliased because it is another magic command.\\n\"}" (:name "stderr" :text "ERROR:root:Invalid alias: The name clear can't be aliased because it is another magic command.\n")) :buffers nil) :inhibited-handlers nil :callbacks ((:shutdown-reply . #f(compiled-function (msg) #<bytecode 0x1240cc1>)))) (:header (message-part "{\"msg_id\":\"e8c6a08d-83b1c9f7d74a3dfb1b80248e\",\"msg_type\":\"stream\",\"username\":\"boy\",\"session\":\"464ef51b-d62e8feaf3cf3b8548035101\",\"date\":\"2019-03-04T11:03:56.326894Z\",\"version\":\"5.3\"}" (:msg_id "e8c6a08d-83b1c9f7d74a3dfb1b80248e" :msg_type :stream :username "boy" :session "464ef51b-d62e8feaf3cf3b8548035101" :date "2019-03-04T11:03:56.326894Z" :version "5.3")) :msg_id "e8c6a08d-83b1c9f7d74a3dfb1b80248e" :msg_type :stream :parent_header (message-part "{\"msg_id\":\"cfd3b78d-1ec6-409b-ba97-41c2cb82c655\",\"msg_type\":\"shutdown_request\",\"version\":\"5.3\",\"username\":\"boy\",\"session\":\"9b3dd7de-5a5f-4878-8643-a179c8a4bee2\",\"date\":\"2019-03-04T12:03:56.171874+01:00\"}" (:msg_id "cfd3b78d-1ec6-409b-ba97-41c2cb82c655" :msg_type :shutdown-request :version "5.3" :username "boy" :session "9b3dd7de-5a5f-4878-8643-a179c8a4bee2" :date "2019-03-04T12:03:56.171874+01:00")) :metadata (message-part "{}" nil) :content (message-part "{\"name\":\"stderr\",\"text\":\"ERROR:root:Invalid alias: The name clear can't be aliased because it is another magic command.\\n\"}" (:name "stderr" :text "ERROR:root:Invalid alias: The name clear can't be aliased because it is another magic command.\n")) :buffers nil)))
  jupyter-handle-message(:iopub #<jupyter-org-client jupyter-org-client> #s(jupyter-request :id "cfd3b78d-1ec6-409b-ba97-41c2cb82c655" :time (23677 1564 171792 52000) :idle-received-p t :last-message (:header (message-part "{\"msg_id\":\"e8c6a08d-83b1c9f7d74a3dfb1b80248e\",\"msg_type\":\"stream\",\"username\":\"boy\",\"session\":\"464ef51b-d62e8feaf3cf3b8548035101\",\"date\":\"2019-03-04T11:03:56.326894Z\",\"version\":\"5.3\"}" (:msg_id "e8c6a08d-83b1c9f7d74a3dfb1b80248e" :msg_type :stream :username "boy" :session "464ef51b-d62e8feaf3cf3b8548035101" :date "2019-03-04T11:03:56.326894Z" :version "5.3")) :msg_id "e8c6a08d-83b1c9f7d74a3dfb1b80248e" :msg_type :stream :parent_header (message-part "{\"msg_id\":\"cfd3b78d-1ec6-409b-ba97-41c2cb82c655\",\"msg_type\":\"shutdown_request\",\"version\":\"5.3\",\"username\":\"boy\",\"session\":\"9b3dd7de-5a5f-4878-8643-a179c8a4bee2\",\"date\":\"2019-03-04T12:03:56.171874+01:00\"}" (:msg_id "cfd3b78d-1ec6-409b-ba97-41c2cb82c655" :msg_type :shutdown-request :version "5.3" :username "boy" :session "9b3dd7de-5a5f-4878-8643-a179c8a4bee2" :date "2019-03-04T12:03:56.171874+01:00")) :metadata (message-part "{}" nil) :content (message-part "{\"name\":\"stderr\",\"text\":\"ERROR:root:Invalid alias: The name clear can't be aliased because it is another magic command.\\n\"}" (:name "stderr" :text "ERROR:root:Invalid alias: The name clear can't be aliased because it is another magic command.\n")) :buffers nil) :inhibited-handlers nil :callbacks ((:shutdown-reply . #f(compiled-function (msg) #<bytecode 0x1240cc1>)))) (:header (message-part "{\"msg_id\":\"e8c6a08d-83b1c9f7d74a3dfb1b80248e\",\"msg_type\":\"stream\",\"username\":\"boy\",\"session\":\"464ef51b-d62e8feaf3cf3b8548035101\",\"date\":\"2019-03-04T11:03:56.326894Z\",\"version\":\"5.3\"}" (:msg_id "e8c6a08d-83b1c9f7d74a3dfb1b80248e" :msg_type :stream :username "boy" :session "464ef51b-d62e8feaf3cf3b8548035101" :date "2019-03-04T11:03:56.326894Z" :version "5.3")) :msg_id "e8c6a08d-83b1c9f7d74a3dfb1b80248e" :msg_type :stream :parent_header (message-part "{\"msg_id\":\"cfd3b78d-1ec6-409b-ba97-41c2cb82c655\",\"msg_type\":\"shutdown_request\",\"version\":\"5.3\",\"username\":\"boy\",\"session\":\"9b3dd7de-5a5f-4878-8643-a179c8a4bee2\",\"date\":\"2019-03-04T12:03:56.171874+01:00\"}" (:msg_id "cfd3b78d-1ec6-409b-ba97-41c2cb82c655" :msg_type :shutdown-request :version "5.3" :username "boy" :session "9b3dd7de-5a5f-4878-8643-a179c8a4bee2" :date "2019-03-04T12:03:56.171874+01:00")) :metadata (message-part "{}" nil) :content (message-part "{\"name\":\"stderr\",\"text\":\"ERROR:root:Invalid alias: The name clear can't be aliased because it is another magic command.\\n\"}" (:name "stderr" :text "ERROR:root:Invalid alias: The name clear can't be aliased because it is another magic command.\n")) :buffers nil))
  jupyter--run-handler-maybe(#<jupyter-org-client jupyter-org-client> :iopub #s(jupyter-request :id "cfd3b78d-1ec6-409b-ba97-41c2cb82c655" :time (23677 1564 171792 52000) :idle-received-p t :last-message (:header (message-part "{\"msg_id\":\"e8c6a08d-83b1c9f7d74a3dfb1b80248e\",\"msg_type\":\"stream\",\"username\":\"boy\",\"session\":\"464ef51b-d62e8feaf3cf3b8548035101\",\"date\":\"2019-03-04T11:03:56.326894Z\",\"version\":\"5.3\"}" (:msg_id "e8c6a08d-83b1c9f7d74a3dfb1b80248e" :msg_type :stream :username "boy" :session "464ef51b-d62e8feaf3cf3b8548035101" :date "2019-03-04T11:03:56.326894Z" :version "5.3")) :msg_id "e8c6a08d-83b1c9f7d74a3dfb1b80248e" :msg_type :stream :parent_header (message-part "{\"msg_id\":\"cfd3b78d-1ec6-409b-ba97-41c2cb82c655\",\"msg_type\":\"shutdown_request\",\"version\":\"5.3\",\"username\":\"boy\",\"session\":\"9b3dd7de-5a5f-4878-8643-a179c8a4bee2\",\"date\":\"2019-03-04T12:03:56.171874+01:00\"}" (:msg_id "cfd3b78d-1ec6-409b-ba97-41c2cb82c655" :msg_type :shutdown-request :version "5.3" :username "boy" :session "9b3dd7de-5a5f-4878-8643-a179c8a4bee2" :date "2019-03-04T12:03:56.171874+01:00")) :metadata (message-part "{}" nil) :content (message-part "{\"name\":\"stderr\",\"text\":\"ERROR:root:Invalid alias: The name clear can't be aliased because it is another magic command.\\n\"}" (:name "stderr" :text "ERROR:root:Invalid alias: The name clear can't be aliased because it is another magic command.\n")) :buffers nil) :inhibited-handlers nil :callbacks ((:shutdown-reply . #f(compiled-function (msg) #<bytecode 0x1240cc1>)))) (:header (message-part "{\"msg_id\":\"e8c6a08d-83b1c9f7d74a3dfb1b80248e\",\"msg_type\":\"stream\",\"username\":\"boy\",\"session\":\"464ef51b-d62e8feaf3cf3b8548035101\",\"date\":\"2019-03-04T11:03:56.326894Z\",\"version\":\"5.3\"}" (:msg_id "e8c6a08d-83b1c9f7d74a3dfb1b80248e" :msg_type :stream :username "boy" :session "464ef51b-d62e8feaf3cf3b8548035101" :date "2019-03-04T11:03:56.326894Z" :version "5.3")) :msg_id "e8c6a08d-83b1c9f7d74a3dfb1b80248e" :msg_type :stream :parent_header (message-part "{\"msg_id\":\"cfd3b78d-1ec6-409b-ba97-41c2cb82c655\",\"msg_type\":\"shutdown_request\",\"version\":\"5.3\",\"username\":\"boy\",\"session\":\"9b3dd7de-5a5f-4878-8643-a179c8a4bee2\",\"date\":\"2019-03-04T12:03:56.171874+01:00\"}" (:msg_id "cfd3b78d-1ec6-409b-ba97-41c2cb82c655" :msg_type :shutdown-request :version "5.3" :username "boy" :session "9b3dd7de-5a5f-4878-8643-a179c8a4bee2" :date "2019-03-04T12:03:56.171874+01:00")) :metadata (message-part "{}" nil) :content (message-part "{\"name\":\"stderr\",\"text\":\"ERROR:root:Invalid alias: The name clear can't be aliased because it is another magic command.\\n\"}" (:name "stderr" :text "ERROR:root:Invalid alias: The name clear can't be aliased because it is another magic command.\n")) :buffers nil))
  #f(compiled-function () #<bytecode 0x1bfbe61>)()
  #f(compiled-function (client channel msg) "Process a message on CLIENT's CHANNEL.\nWhen a message is received from the kernel, the\n`jupyter-handle-message' method is called on the client. The\nclient method runs any callbacks for the message and possibly\nruns the client handler for the channel the message was received\non. The channel's `jupyter-handle-message' method will then pass\nthe message to the appropriate message handler based on the\nmessage type.\n\nSo when a message is received from the kernel the following steps\nare taken:\n\n- `jupyter-handle-message' (client)\n   - Run callbacks for message\n   - Possibly run channel handlers\n     - `jupyter-handle-message' (channel)\n       - Based on message type, dispatch to\n         `jupyter-handle-execute-result',\n         `jupyter-handle-kernel-info-reply', ...\n   - Remove request from client request table when idle message is received" #<bytecode 0x2318a99>)(#<jupyter-org-client jupyter-org-client> :iopub (:header (message-part "{\"msg_id\":\"e8c6a08d-83b1c9f7d74a3dfb1b80248e\",\"msg_type\":\"stream\",\"username\":\"boy\",\"session\":\"464ef51b-d62e8feaf3cf3b8548035101\",\"date\":\"2019-03-04T11:03:56.326894Z\",\"version\":\"5.3\"}" (:msg_id "e8c6a08d-83b1c9f7d74a3dfb1b80248e" :msg_type :stream :username "boy" :session "464ef51b-d62e8feaf3cf3b8548035101" :date "2019-03-04T11:03:56.326894Z" :version "5.3")) :msg_id "e8c6a08d-83b1c9f7d74a3dfb1b80248e" :msg_type :stream :parent_header (message-part "{\"msg_id\":\"cfd3b78d-1ec6-409b-ba97-41c2cb82c655\",\"msg_type\":\"shutdown_request\",\"version\":\"5.3\",\"username\":\"boy\",\"session\":\"9b3dd7de-5a5f-4878-8643-a179c8a4bee2\",\"date\":\"2019-03-04T12:03:56.171874+01:00\"}" (:msg_id "cfd3b78d-1ec6-409b-ba97-41c2cb82c655" :msg_type :shutdown-request :version "5.3" :username "boy" :session "9b3dd7de-5a5f-4878-8643-a179c8a4bee2" :date "2019-03-04T12:03:56.171874+01:00")) :metadata (message-part "{}" nil) :content (message-part "{\"name\":\"stderr\",\"text\":\"ERROR:root:Invalid alias: The name clear can't be aliased because it is another magic command.\\n\"}" (:name "stderr" :text "ERROR:root:Invalid alias: The name clear can't be aliased because it is another magic command.\n")) :buffers nil))
  apply(#f(compiled-function (client channel msg) "Process a message on CLIENT's CHANNEL.\nWhen a message is received from the kernel, the\n`jupyter-handle-message' method is called on the client. The\nclient method runs any callbacks for the message and possibly\nruns the client handler for the channel the message was received\non. The channel's `jupyter-handle-message' method will then pass\nthe message to the appropriate message handler based on the\nmessage type.\n\nSo when a message is received from the kernel the following steps\nare taken:\n\n- `jupyter-handle-message' (client)\n   - Run callbacks for message\n   - Possibly run channel handlers\n     - `jupyter-handle-message' (channel)\n       - Based on message type, dispatch to\n         `jupyter-handle-execute-result',\n         `jupyter-handle-kernel-info-reply', ...\n   - Remove request from client request table when idle message is received" #<bytecode 0x2318a99>) #<jupyter-org-client jupyter-org-client> (:iopub (:header (message-part "{\"msg_id\":\"e8c6a08d-83b1c9f7d74a3dfb1b80248e\",\"msg_type\":\"stream\",\"username\":\"boy\",\"session\":\"464ef51b-d62e8feaf3cf3b8548035101\",\"date\":\"2019-03-04T11:03:56.326894Z\",\"version\":\"5.3\"}" (:msg_id "e8c6a08d-83b1c9f7d74a3dfb1b80248e" :msg_type :stream :username "boy" :session "464ef51b-d62e8feaf3cf3b8548035101" :date "2019-03-04T11:03:56.326894Z" :version "5.3")) :msg_id "e8c6a08d-83b1c9f7d74a3dfb1b80248e" :msg_type :stream :parent_header (message-part "{\"msg_id\":\"cfd3b78d-1ec6-409b-ba97-41c2cb82c655\",\"msg_type\":\"shutdown_request\",\"version\":\"5.3\",\"username\":\"boy\",\"session\":\"9b3dd7de-5a5f-4878-8643-a179c8a4bee2\",\"date\":\"2019-03-04T12:03:56.171874+01:00\"}" (:msg_id "cfd3b78d-1ec6-409b-ba97-41c2cb82c655" :msg_type :shutdown-request :version "5.3" :username "boy" :session "9b3dd7de-5a5f-4878-8643-a179c8a4bee2" :date "2019-03-04T12:03:56.171874+01:00")) :metadata (message-part "{}" nil) :content (message-part "{\"name\":\"stderr\",\"text\":\"ERROR:root:Invalid alias: The name clear can't be aliased because it is another magic command.\\n\"}" (:name "stderr" :text "ERROR:root:Invalid alias: The name clear can't be aliased because it is another magic command.\n")) :buffers nil)))
  jupyter-handle-message(#<jupyter-org-client jupyter-org-client> :iopub (:header (message-part "{\"msg_id\":\"e8c6a08d-83b1c9f7d74a3dfb1b80248e\",\"msg_type\":\"stream\",\"username\":\"boy\",\"session\":\"464ef51b-d62e8feaf3cf3b8548035101\",\"date\":\"2019-03-04T11:03:56.326894Z\",\"version\":\"5.3\"}" (:msg_id "e8c6a08d-83b1c9f7d74a3dfb1b80248e" :msg_type :stream :username "boy" :session "464ef51b-d62e8feaf3cf3b8548035101" :date "2019-03-04T11:03:56.326894Z" :version "5.3")) :msg_id "e8c6a08d-83b1c9f7d74a3dfb1b80248e" :msg_type :stream :parent_header (message-part "{\"msg_id\":\"cfd3b78d-1ec6-409b-ba97-41c2cb82c655\",\"msg_type\":\"shutdown_request\",\"version\":\"5.3\",\"username\":\"boy\",\"session\":\"9b3dd7de-5a5f-4878-8643-a179c8a4bee2\",\"date\":\"2019-03-04T12:03:56.171874+01:00\"}" (:msg_id "cfd3b78d-1ec6-409b-ba97-41c2cb82c655" :msg_type :shutdown-request :version "5.3" :username "boy" :session "9b3dd7de-5a5f-4878-8643-a179c8a4bee2" :date "2019-03-04T12:03:56.171874+01:00")) :metadata (message-part "{}" nil) :content (message-part "{\"name\":\"stderr\",\"text\":\"ERROR:root:Invalid alias: The name clear can't be aliased because it is another magic command.\\n\"}" (:name "stderr" :text "ERROR:root:Invalid alias: The name clear can't be aliased because it is another magic command.\n")) :buffers nil))
  apply(jupyter-handle-message (#<jupyter-org-client jupyter-org-client> :iopub (:header (message-part "{\"msg_id\":\"e8c6a08d-83b1c9f7d74a3dfb1b80248e\",\"msg_type\":\"stream\",\"username\":\"boy\",\"session\":\"464ef51b-d62e8feaf3cf3b8548035101\",\"date\":\"2019-03-04T11:03:56.326894Z\",\"version\":\"5.3\"}" (:msg_id "e8c6a08d-83b1c9f7d74a3dfb1b80248e" :msg_type :stream :username "boy" :session "464ef51b-d62e8feaf3cf3b8548035101" :date "2019-03-04T11:03:56.326894Z" :version "5.3")) :msg_id "e8c6a08d-83b1c9f7d74a3dfb1b80248e" :msg_type :stream :parent_header (message-part "{\"msg_id\":\"cfd3b78d-1ec6-409b-ba97-41c2cb82c655\",\"msg_type\":\"shutdown_request\",\"version\":\"5.3\",\"username\":\"boy\",\"session\":\"9b3dd7de-5a5f-4878-8643-a179c8a4bee2\",\"date\":\"2019-03-04T12:03:56.171874+01:00\"}" (:msg_id "cfd3b78d-1ec6-409b-ba97-41c2cb82c655" :msg_type :shutdown-request :version "5.3" :username "boy" :session "9b3dd7de-5a5f-4878-8643-a179c8a4bee2" :date "2019-03-04T12:03:56.171874+01:00")) :metadata (message-part "{}" nil) :content (message-part "{\"name\":\"stderr\",\"text\":\"ERROR:root:Invalid alias: The name clear can't be aliased because it is another magic command.\\n\"}" (:name "stderr" :text "ERROR:root:Invalid alias: The name clear can't be aliased because it is another magic command.\n")) :buffers nil)))
  timer-event-handler([t 23677 1564 328288 nil jupyter-handle-message (#<jupyter-org-client jupyter-org-client> :iopub (:header (message-part "{\"msg_id\":\"e8c6a08d-83b1c9f7d74a3dfb1b80248e\",\"msg_type\":\"stream\",\"username\":\"boy\",\"session\":\"464ef51b-d62e8feaf3cf3b8548035101\",\"date\":\"2019-03-04T11:03:56.326894Z\",\"version\":\"5.3\"}" (:msg_id "e8c6a08d-83b1c9f7d74a3dfb1b80248e" :msg_type :stream :username "boy" :session "464ef51b-d62e8feaf3cf3b8548035101" :date "2019-03-04T11:03:56.326894Z" :version "5.3")) :msg_id "e8c6a08d-83b1c9f7d74a3dfb1b80248e" :msg_type :stream :parent_header (message-part "{\"msg_id\":\"cfd3b78d-1ec6-409b-ba97-41c2cb82c655\",\"msg_type\":\"shutdown_request\",\"version\":\"5.3\",\"username\":\"boy\",\"session\":\"9b3dd7de-5a5f-4878-8643-a179c8a4bee2\",\"date\":\"2019-03-04T12:03:56.171874+01:00\"}" (:msg_id "cfd3b78d-1ec6-409b-ba97-41c2cb82c655" :msg_type :shutdown-request :version "5.3" :username "boy" :session "9b3dd7de-5a5f-4878-8643-a179c8a4bee2" :date "2019-03-04T12:03:56.171874+01:00")) :metadata (message-part "{}" nil) :content (message-part "{\"name\":\"stderr\",\"text\":\"ERROR:root:Invalid alias: The name clear can't be aliased because it is another magic command.\\n\"}" (:name "stderr" :text "ERROR:root:Invalid alias: The name clear can't be aliased because it is another magic command.\n")) :buffers nil)) nil 278000])
  accept-process-output(nil 0.0001)
  jupyter-repl-restart-kernel(nil nil)
  funcall-interactively(jupyter-repl-restart-kernel nil nil)
  call-interactively(jupyter-repl-restart-kernel)
  jupyter-org--call-with-src-block-client(jupyter-repl-restart-kernel)
  jupyter-repl-restart-kernel()
  funcall-interactively(jupyter-repl-restart-kernel)
  call-interactively(jupyter-repl-restart-kernel nil nil)
  command-execute(jupyter-repl-restart-kernel)
```

jupyter-org-client.el
UndeadKernel: I was not aware that the new standard was to use lower case for all org keywords.
As long as it's consistent, I don't mind then the lower case keywords.
I wouldn't mind then using `org-element-interpret-data` and letting org-mode choose the case.

What would you prefer?

jupyter-org-extensions.el
UndeadKernel: Changed to use `jupyter-completing-read-kernelspec`.

jupyter-org-extensions.el
UndeadKernel: When it comes to using org mode to replicate jupyter notebooks, I'd argue that moving source blocks between boundaries is desired. This is because in jupyter notebooks there is no such concept as a subtree or a boundary.
The current behavior would be more in-line with how notebooks operate.

I also find this functionality more useful. One workflow I use often is to clone a source block (it's written below the current lbock) and then move the new block to a new location, often far from the original block.

What is your take on this?

jupyter-org-extensions.el
UndeadKernel: Of course...

jupyter-org-extensions.el
UndeadKernel: Much better, yes.

jupyter-org-extensions.el
UndeadKernel: I think the logic behind this idea would be rather ugly to implement:
The idea is to be able to merge blocks next to each other as well as those that have org elements in between. For example, if there is text in between two blocks, it's still desirable to be able to merge them together without touching the org elements in the middle.

jupyter-org-extensions.el
UndeadKernel: Good point. I've switched to `delete-and-extract-region`.
I've also simplified this function (i.e. `jupyter-org-move-src-block`).

jupyter-org-extensions.el
dzop: Yes your right it doesn't make sense to delete the intermediate text since there could also be other stuff associated with the text, like overlays, that would be removed. I was thinking since the end result would be the same, it wouldn't matter that we messed with the intermediate text.

Below is a different take on the idea that just deletes the source block after the current block. I believe it is simpler than the `org-element` solution since it only deletes the contents of the source block after the current one instead of also deleting the current one.

```emacs-lisp
(defun jupyter-org-merge-blocks ()
  "Merge the current block with the next block."
  (interactive)
  (unless (org-in-src-block-p)
    (error "Not in a source block"))
  (let ((asrc (org-element-context)))
    (org-babel-remove-result)
    (org-babel-next-src-block)
    (let* ((bsrc (prog1 (org-element-context)
                   (org-babel-remove-result)))
           (bsrc-beg (set-marker
                      (make-marker)
                      (org-element-property :begin bsrc)))
           (bsrc-end (set-marker
                      (make-marker)
                      (jupyter-org-element-end-before-blanks bsrc))))
      (goto-char (jupyter-org-element-end-before-blanks asrc))
      (forward-line -1)
      (insert
       (delete-and-extract-region
        (save-excursion
          (goto-char (jupyter-org-element-begin-after-affiliated bsrc))
          (forward-line 1)
          (point))
        (save-excursion
          (goto-char bsrc-end)
          (forward-line -1)
          (point))))
      (delete-region bsrc-beg bsrc-end)
      (set-marker bsrc-beg nil)
      (set-marker bsrc-end nil))))
```

jupyter-org-client.el
dzop: OK, I figured out the error. It has something to do with the `:shutdown-request` causing a `:stream` message to be emitted by the kernel, and the `jupyter-stream-handler` of the REPL client not handling the case when no input cell is associated with a request. Thanks for catching this, I don't think it matters that we handle any messages other than `:shutdown-reply` for a `:shutdown-request` since there would never be an association with a REPL input cell.

This normally doesn't happen for me because I have never received the error

`ERROR:root:Invalid alias: The name clear can't be aliased because it is another magic command.`

But it looks to be an issue with IPython 7.3.0 https://stackoverflow.com/questions/54971047/spyder-reports-invalid-alias-when-running-any-script, is that what you are on?

jupyter-org-extensions.el
dzop: OK, if its similar to the Jupyter notebook. I never really used the notebook features that much, but if people are used to that workflow then it doesn't make sense to not allow it.

jupyter-org-client.el
dzop: You shouldn't get any errors now. I just ignored any other messages that are not a `:shutdown-reply` when restarting the kernel.

jupyter-org-extensions.el
```diff
@@ -0,0 +1,394 @@
...
+
+;;;###autoload
+(defun jupyter-org-jump-to-visible-block ()
+  "Jump to a visible src block with avy."
+  (interactive)
+  (avy-with jupyter-org-jump-to-block
+            (avy-jump "#\\+begin_src" :beg (point-min) :end (point-max))))
+
+;;;###autoload
+(defun jupyter-org-edit-header ()b
```
dzop: Stray `b` at the end here.

jupyter-org-extensions.el
```diff
@@ -0,0 +1,394 @@
...
+           (header-end (save-excursion (goto-char header-start)
+                                       (line-end-position))))
+      (setf (buffer-substring header-start header-end)
+            (read-string "Header: "
+                         (buffer-substring header-start header-end))))))
+
+(defun jupyter-org-select-block-and-results ()
+  "Return the region that covers a source block and its results (if any).
+
+Return the region as (REGION_START . REGION_END)"
```
dzop: `BEGIN` seems better here since `org-element` already uses `:begin` for the beginning of a source block. Also we can just say `(BEGIN . END)` here since the region is already mentioned.

Also I think `contains` sounds more correct here than `covers` since the text that is the source block is contained in the buffer region between `(BEGIN . END)`. WDYT?

jupyter-org-extensions.el
```diff
@@ -0,0 +1,394 @@
...
+    (error "Not in a source block"))
+  (let* ((src (org-element-context))
+         (results-start (org-babel-where-is-src-block-result))
+         (results-end
+          (when results-start
+            (save-excursion
+              (goto-char results-start)
+              (goto-char (org-babel-result-end))
+              (point)))))
+    ;; return the region as (REGION_START . REGION_END)
```
dzop: Isn't this comment a little redundant since we already mention what the return value is in the doc string?

jupyter-org-client.el
UndeadKernel: That was it. The kernel is restarting as expected and all the extension functions that rely on it are working fine.

jupyter-org-extensions.el
UndeadKernel: Your proposal looks good.
I've modified to delete one extra line that was leftover and to move to the end of the last line.

jupyter-org-extensions.el
```diff
@@ -0,0 +1,401 @@
...
+;; Functions that extend the functionality of Org mode to interact with
+;; jupyter source blocks.
+
+;;; Code:
+
+(require 'org)
+(require 'jupyter-org-client)
+
+(declare-function org-babel-jupyter-initiate-session "ob-jupyter" (&optional session params))
+(declare-function avy-with nil)
```
dzop: Can you add the package where these come from, e.g. using `(declare-function avy-with "ext:avy")`

jupyter-org-extensions.el
```diff
@@ -0,0 +1,401 @@
...
+
+If a new block is created, use the same language, switches and parameters.
+With prefix arg NEW, always insert new cell."
+  (interactive "P")
+  (unless (org-in-src-block-p)
+      (error "Not in a source block"))
+  (org-babel-execute-src-block)
+  ;; add a new src block if there is no next one to jump to
+  (unless (or new (ignore-errors (org-babel-next-src-block)))
+    ;; FIXME: without this, the new block is added as part of the results of the last block
```
dzop: Why don't we add the new source block first and then execute the current one?

jupyter-org-extensions.el
```diff
@@ -0,0 +1,401 @@
...
+        (setq block (concat block "\n")))
+      (if below
+          ;; if below, move past the next source block or its result
+          (let ((next-src-block-head (org-babel-where-is-src-block-head)))
+            (if next-src-block-head
+                (goto-char next-src-block-head)
+              (org-babel-next-src-block))
+            (let ((next-src-block (org-element-context))
+                  (next-results-start (org-babel-where-is-src-block-result)))
+              (if next-results-start
```
dzop: What about `(if (null ...` to get rid of the `progn`?

jupyter-org-extensions.el
UndeadKernel: I feel bad for not seeing that solution before :)

jupyter-org-extensions.el
UndeadKernel: Sure, can be done.

jupyter-org-extensions.el
```diff
@@ -0,0 +1,411 @@
...
+  (jupyter-org-with-src-block-client
+   (jupyter-repl-restart-kernel))
+  (org-babel-execute-buffer))
+
+;;;###autoload
+(defun jupyter-org-jump-to-block (&optional N)
+  "Jump to a block in the buffer.
+If narrowing is in effect, only a block in the narrowed region.
+Use a numeric prefix N to specify how many lines of context to use.
+Defaults to 3."
```
dzop: I think it would be better to have this as a customizable variable. Something like
`jupyter-org-jump-to-block-context-lines` with doc string 
``Default number of lines to use for context in `jupyter-org-jump-to-block'``.

Should be good to go for merging after this!

UndeadKernel: Not a bad idea.

jupyter-org-extensions.el
```diff
@@ -158,16 +165,17 @@ With prefix arg NEW, always insert new cell."
...
+If narrowing is in effect, jump to a block in the narrowed region.
+Use a numeric prefix CONTEXT to specify how many lines of context to use.
 Defaults to 3."
   (interactive "p")
   (unless (require 'ivy nil t)
     (error "Package `ivy' not installed"))
   (let ((p '()))
-    (when (= 1 N) (setq N 3))
+    (when (= 1 context)
+      (setq context jupyter-org-jump-to-block-context-lines))
```
dzop: Actually we should set this in the interactive spec with something like

```emacs-lisp
(interactive
 (list (if context (prefix-numeric-value context)
         jupyter-org-jump-to-block-context-lines)))
```

Then here we would have

```emacs-lisp
(or context (setq context jupyter-org-jump-to-block-context-lines))
```

The way we have it now would actually fail if `context` is `nil` since you can't compare `nil` with a number.

UndeadKernel: I don't think this is quite needed:
The interactive declaration I'm using [i.e., `(interactive "p") ] already does the equivalent of `prefix-numeric-value`. In theory, `context` can only be a number.

dzop: I am thinking more of the non-interactive case. You are right that the `interactive` spec doesn't need to change, but the function itself specifies that `context` is optional which means that it can be `nil` if called directly. For example calling `(jupyter-org-jump-to-block)` directly is valid and in this case, `context` will be `nil`. So really the issue is that a `nil` `context` needs to be handled. Changing the `interactive` spec really has nothing to do with the issue, sorry for the noise.

Also I messed up the interactive spec, it should be

```emacs-lisp
(interactive
 (list (if current-prefix-arg (prefix-numeric-value current-prefix-arg)
         jupyter-org-jump-to-block-context-lines)))
```

UndeadKernel: How about having the following change instead?
```lisp
(when (or (= 1 context) (not context))
      (setq context jupyter-org-jump-to-block-context-lines))
```

dzop: OK, I think this is where changing the interactive spec has benefits since it doesn't rely on needing to handle the special case of `context` being 1. With this change, if I where to evaluate 
`(jupyter-org-jump-to-block 1)` it would actually use `jupyter-org-jump-to-block-context-lines` for `context` instead of 1 line. If we change the interactive spec, then we don't need to handle this special case in the actual function.

In the function body, we don't know if the function has been called interactively or not, we could use `called-interactively-p` to check but changing the interactive spec is the more standard approach in this case.

UndeadKernel: It's intentional that `(jupyter-org-jump-to-block 1)` sets the context to `jupyter-org-jump-to-block-context-lines` because there is no point in the context being `1`. If that were the case, only the `#+SRC_BLOCK` line would be shown without a proper context.
On the other hand, I'd in theory have to check if the context is below 1 and change it to `jupyter-org-jump-to-block-context-lines`. I'm therefore thinking that this change would also make sense:
```lisp
(when (or (< context 1) (not context))
      (setq context jupyter-org-jump-to-block-context-lines))
```
What do you think about this?

dzop: > there is no point in the context being 1. If that were the case, only the #+SRC_BLOCK line would be shown without a proper context.

In that case, should the context mean the number of actual source lines?



If we keep with a `context` of 1 actually meaning `jupyter-org-jump-to-block-context-lines`, then if I do `(jupyter-org-jump-to-block 2)`, it will actually use a number of lines less than the default value of `jupyter-org-jump-to-block-context-lines` (assuming it is 3). Should this be possible?

I agree we should handle the case of the context being less than 1.

UndeadKernel: I've integrated your suggestions into what I believe is the better solution: a custom interactive spec and a bounds check. The CONTEXT variable now means "lines of code after the header of the source block".

dzop:
Thanks!

PR #56, opened 2019-03-12
Handle ANSI escape sequences in `org-mode` src-block results

### LINKED ISSUE #55: Colour in Results block?
Currently, the code shown in the output block is black and white whereas an associated repl session has the proper colours. Is there a way around this? 

Here's an example:

Screen Shot 2019-03-11 at 11 18 36 AM

dzop:
This pull request causes ANSI escape sequences in `org-mode` example blocks or fixed-width (lines that start with `: `) elements to be colored and the actual escape sequences are rendered invisible.

There are issues with this though, since now the escape sequences exist in the buffer instead of being removed before insertion. This means that when exporting to other formats those escape sequences will still be in the exported output.

One way to address this is to hook into `org-mode`'s export functionality and remove the escape sequences in the `example-block` and `fixed-width` elements that `org-mode` processes during export. This should only be done for `example-block` and `fixed-width` elements that are part of the results of Jupyter source blocks since there may be situations where exporting elements with ANSI escape sequences is wanted.

In fact, the features of this pull request should probably only be applied to Jupyter source block results. The current implementation applies to all ``example-block`` or ``fixed-width`` elements in the buffer whenever `jupyter-org-interaction-mode` is enabled.

Closes #55.

### DIFF
--- a/jupyter-mime.el
+++ b/jupyter-mime.el
@@ -184,6 +184,69 @@ function."
         (kill-buffer)))
     (funcall restore-mode)))
 
+;;; Special handling of ANSI sequences
+
+(defun jupyter-ansi-color-apply-on-region (begin end)
+  "`ansi-color-apply-on-region' with Jupyter specific modifications.
+In particular, does not delete escape sequences between BEGIN and
+END from the buffer. Instead, an invisible text property with a
+value of t is added to render the escape sequences invisible.
+Also, the `ansi-color-apply-face-function' is hard-coded to a
+custom function that prepends to the face property of the text
+and also sets the font-lock-face to the prepended face.
+
+For convenience, a jupyter-invisible property is also added with
+a value of t. This is mainly for modes like `org-mode' which
+strip invisible properties during fontification. In such cases,
+the jupyter-invisible property can act as an alias to the
+invisible property by adding it to `char-property-alias-alist'."
+  (let ((codes (car ansi-color-context-region))
+        (start-marker (or (cadr ansi-color-context-region)
+                          (copy-marker begin)))
+        (end-marker (copy-marker end))
+        (ansi-color-apply-face-function
+         (lambda (beg end face)
+           (when face
+             (setq face (list face))
+             (font-lock-prepend-text-property beg end 'face face)
+             (put-text-property beg end 'font-lock-face face)))))
+    (save-excursion
+      (goto-char start-marker)
+      ;; Find the next escape sequence.
+      (while (re-search-forward ansi-color-control-seq-regexp end-marker t)
+        ;; Remove escape sequence.
+        (let ((esc-seq (prog1 (buffer-substring-no-properties
+                               (match-beginning 0) (point))
+                         ;; FIXME: Not removing escape sequences adds in a lot
+                         ;; of invisible characters that slows down Emacs on
+                         ;; large ANSI coded regions and seems mostly related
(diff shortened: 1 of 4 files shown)


dzop:
There are also some inefficiencies here since the handling of ANSI escapes happens during font-lock which means that the faces can be re-computed for the region many times during re-display.

dzop:
To hook into the export process we can add a function to `org-export-before-parsing-hook`. Then, depending on the backend, we can loop over the source blocks in the buffer with `org-babel-map-src-blocks` and strip the escape sequences for every block whose results have example or fixed-width blocks. The functions in `org-export-before-parsing-hook` run on a copy of the buffer being exported so it should be OK to modify the buffer.

ssfrr:
Is this merged into the version in MELPA?

Currently (version 20190917.1624) I see this:
image

dzop:
Yes these changes should be on the MELPA version. I'm seeing

Screen Shot 2019-09-21 at 10 02 34 PM

You need to have `jupyter-org-interaction-mode` enabled for the ANSI escape sequences to be colorized. If you already have that mode enabled, when you evaluate

```emacs-lisp
(cl-find-if (lambda (x) (eq (car x) 'jupyter-org-font-lock-ansi-escapes))
            org-font-lock-keywords)
```
in an `org-mode` buffer does it return something?

ssfrr:
Thanks for the help. I think it's enabled (when I run `jupyter-org-interaction-mode` it tells me that it disables it, and running it again re-enables it).

That code returns `(jupyter-org-font-lock-ansi-escapes)`.

If it's relevant this is emacs 26.3, with org version 20190923.

dzop:
Ok could you open up a new issue and in that new issue show the output of the following code

```emacs-lisp
(with-temp-buffer
  (org-mode)
  (insert ": [31mhello[39m")
  (font-lock-ensure)
  (text-properties-at 8))
```

You should get

```
(font-lock-face ((foreground-color . "#dc322f")) 
  jupyter-ansi t 
  face ((foreground-color . "#dc322f") org-code))
```

About the Emacs version, we run tests against the latest `org-mode` and against Emacs 26.3 so it shouldn't be a problem.

Can you also maybe try to reproduce the issue in an `emacs -Q` session with just `org-mode` and `emacs-jupyter` loaded, e.g. by putting the below in a file `test-ansi.el`

```emacs-lisp
(push "~/.emacs.d/core/site-lisp/jupyter" load-path)
(push "~/.emacs.d/el-get/simple-httpd/" load-path)
(push "~/.emacs.d/el-get/websocket/" load-path)
(push "~/.emacs.d/el-get/org-mode/lisp/" load-path)
(require 'org)
(require 'jupyter-org-client)
(with-temp-buffer
  (org-mode)
  (insert ": [31mhello[39m")
  (font-lock-ensure)
  (prin1 (text-properties-at 8)))
```

(changing the paths to match those on your system of course) and then run the script like
`emacs -Q -batch -l ansi-test.el`. Note you will have replace those question mark characters by typing `C-q C-[` after copy-pasting to insert the `^[` escape character.

You shouldn't see the `jupyter-ansi` text property or `(foreground-color . "#dc322f")` in the printed output if there is a problem. If you do see those, then the issue most likely lies within your setup for `org-mode`.

stardiviner:
I think this handling function should be more general. And should be merged to Org Mode upstream source code. WDYT?

PR #67, opened 2019-03-22
quickfix in jupyter-kernel-manager--cleanup

### LINKED ISSUE #65: Fail to cleanup kernel manager: file-exists-p passed nil
Occasionally I hit `Wrong type argument: stringp, nil` when shutting down Emacs.

I traced it to `jupyter-kernel-manager--cleanup` passing `nil` to `file-exists-p`.

Unsure what the upstream cause is, but here's my quick-fix to guard against this:

Line 77 of `jupyter-kernel-manager.el`:

```
(when (and conn-file (file-exists-p conn-file))
      (delete-file conn-file))
```

dangirsh:
Fixes #65


dzop:
Thanks.

Issue #72, opened 2019-03-29
[suggestion] easier remote kernel usage


benneti:
What do you think of the idea to provide a simpler version of the `:session /ssh:user@remote:/run/.../jupyter/kernel-....json` with just `:session /ssh:user@remote` that starts the kernel on its own and figures out the filename.

I see that there might be an issues as it does not know the kernels on the remote from the beginning but I also feel like there should be a better way then starting the kernel on the remote and then pasting the name in the org file.

dzop:
Yes it is possible to get the kernel information on the remote host, `jupyter-available-kernelspecs` already handles remote directories. So it would just be of modifying `org-babel-jupyter-initiate-session-by-key` to be aware of this format for `:session` and probably `jupyter-start-new-kernel` to consider remote file names.

The design of the remote kernel stuff was more focused on how to connect to an already running kernel, e.g. on some computing cluster as opposed to starting a kernel on the remote host.

benneti:
Thanks that sounds promising, I'll have a look at it.

benneti:
I think `jupyter-available-kernelspecs` needs a small change, too.
So what I thought of so far would that when one wants well defined behavior and different sessions on remote hosts one first needs to differentiate sessions ending with ".json" and not.
What I mean I hope can be seen in the code (that would need to distributed to the functions you talked about before) and the code is only meant to see whether this is how you would imagine it
``` emacs-lisp
;; connection file provided
(setq default-directory "/ssh:hostname:connectme.json")
;; no connection file but a named remote session
(setq default-directory "/ssh:hostname:sessionname")
;; check wether to use ssh or not
(when (string-prefix-p "/ssh:" default-directory)
  ;; check wether to use connection file
  (when (string-suffix-p ".json" default-directory)
  ;; old beviour
    )
  ;; This is the replacement necessary in jupyter-available-kernelspecs to get the kernelspec
  (let* ((host (nth 1 (split-string default-directory ":")))
         (session (nth 2 (split-string default-directory ":"))))
    (shell-command-to-string (concat "ssh " host " jupyter kernelspec list")))
  ;; then create session with necessary kernel and session id on host (org-babel-jupyter-initiate-session-by-key)
  )
```

dzop:
[...]

PR #75, opened 2019-03-30
[WIP] Improve remote kernel capabilities

benneti:
See issure https://github.com/dzop/emacs-jupyter/issues/72.
Sofar this only changes `jupyter-available-kernelspecs` to use ssh to get the kernelspec folders of the remote host.
Planned to be done in addition:
Automatic remote sessions in org blocks

### DIFF
--- a/jupyter-kernelspec.el
+++ b/jupyter-kernelspec.el
@@ -69,10 +69,9 @@ information of the host as a prefix."
   (let ((host (or (file-remote-p default-directory) "local")))
     (or (and (not refresh) (gethash host jupyter--kernelspecs))
         (let ((spec-list
-               (cdr
-                (split-string
-                 (shell-command-to-string "jupyter kernelspec list")
-                 "\n" 'omitnull "[ \t]+")))
+               (cdr (split-string
+                     (jupyter-command "kernelspec" "list")
+                     "\n" 'omitnull "[ \t]+")))
               (get-kernelspec
                (lambda (s)
                  (cl-destructuring-bind (kernel dir)
--- a/jupyter-base.el
+++ b/jupyter-base.el
@@ -43,6 +43,17 @@
 (declare-function tramp-file-name-host "tramp")
 (declare-function jupyter-message-content "jupyter-messages" (msg))
 
+;;; Call the Jupyter command
+
+(defun jupyter-command (&rest args)
+  "Run a Jupyter shell command synchronously, return its output.
+The shell command run is
+
+    jupyter ARGS..."
+  (with-temp-buffer
+    (apply #'process-file "jupyter" nil t nil args)
+    (string-trim-right (buffer-string))))
+
 ;;; Custom variables
 
 (defcustom jupyter-include-other-output nil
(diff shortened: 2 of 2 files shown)


jupyter-kernelspec.el
```diff
@@ -71,7 +71,12 @@ information of the host as a prefix."
         (let ((spec-list
                (cdr
                 (split-string
-                 (shell-command-to-string "jupyter kernelspec list")
+                 (if (string-prefix-p "/ssh:" host)
```
dzop: Actually we don't need this change, the intended usage for remote directories is something like

```emacs-lisp
(let ((default-directory "/ssh:ec2:~"))
  (jupyter-available-kernelspecs))
```

See https://www.gnu.org/software/emacs/manual/html_node/elisp/Magic-File-Names.html, and `process-file` which is what is called by `shell-command-to-string`.

benneti: oh sorry, I didn't know this and it didn't work for me either (as my local environment and the remote do not have the same login shell and I think `(shell-command-to-string "jupyter kernelspec list")` does not like this).
My local shell is fish and the remote is bash and fish is not even installed there so running 
```
(let ((default-directory "/ssh:fau2:~"))
    (shell-command-to-string "jupyter kernelspec list")))
```
results in
`: /bin/sh: 23: /usr/bin/fish: not found`.
Changing it to
```
(let ((default-directory "/ssh:fau2:~")
        (shell-file-name "/bin/bash"))
    (shell-command-to-string "jupyter kernelspec list")))
```
works.
So now I believe it is an issue with `simple.el` as it uses the `shell-file-name` of the local system even for remote calls.

dzop: Yes it does look like `shell-command-to-string` is messed up and doesn't handle remote directories properly. I replaced it with a new function `jupyter-command`. If you rebase against the latest master, this problem should be fixed.

benneti:
You are awesome 👍

benneti:
Sorry to chime in again but the new function is still not the optimal solution, as calling the function on the remote via `/bin/sh` does not load the path set by the user.  I do not think it is possible to use a jupyter binary in the user directory this way.

dzop:
Why not set the `PATH` on the remote to find the `jupyter` executable? This is more of an issue with how things are setup on your remote. You can update `PATH` in `~/.profile` so that `/bin/sh` can find it.

With regards to `shell-command-to-string` I don't know what the best way to approach this is, but I don't think relying on `printenv` is the way to go. The reason why `/bin/sh` is the default is because that is guaranteed to exist on all Unix based systems (see http://pubs.opengroup.org/onlinepubs/007904975/utilities/sh.html), I'm not so sure about `printenv`. If I had to guess, `shell-command-to-string` should probably use `process-file-shell-command`.

benneti:
Ah in principle there is no problem in setting the path in the profile that is what I will do now. 
The reason for me until know not to do it, was that when I do it in the bash/fish configuration I can set the PATH and other environment variables depending on hostname username or the system it runs on such that it is easier to share the same configuration across many systems.

You have convinced me (not that it was necessary it is your project after all) that for emacs-jupyter your solution is best. I still think `shell-command-to-string` should use the configured shell of the host as the name of the function suggests that shell commands including aliases and such should work.

Thank you again for taking the time to answer!

dzop:
You can use a different `SHELL` if you really wanted to, e.g. by specifying a new method in `tramp-methods` that overrides the `tramp-remote-shell` variable. Not with `shell-command-to-string` obviously, but anything that uses `process-file` will use the `SHELL` in `tramp-remote-shell`.

benneti:
FYI: `shell-command-to-string` works as one would expect in emacs master, so hopefully in the next major emacs version, too.

PR #83, opened 2019-04-06
Set show-trailing-whitespace to nil for jupyter buffers.

### LINKED ISSUE #79: [Aesthetics] Whitespace cleanup after jupyter-repl-insert-banner?
This is pretty much a non issue, but I have `show-trailing-whitespace` set to true, and because the conda Python banner has a trailing whitespace, the jupyter REPL buffer at my end has a trailing whitespace marked, which looks ugly.

I currently solve this issue with the following advice:

```el
  (defun jupyter-banner-cleanup-whitespace (orig-func &rest args)
    "The function jupyter-repl-insert-banner is responsible for inserting
a language banner into the beginning of the REPL buffer. The banner is the
first thing printed when we start a language interpreter.
The issue is that the Python banner contains trailing whitespaces, and so
may look ugly if the variable `show-trailing-whitespace' is set to t."
    (apply orig-func args)
    (whitespace-cleanup))

  (advice-add 'jupyter-repl-insert-banner :around #'jupyter-banner-cleanup-whitespace)
```

Would this merit a patch directly at emacs-jupyter?

dangom:
Closes  #79

### DIFF
--- a/jupyter-base.el
+++ b/jupyter-base.el
@@ -296,6 +296,9 @@ See `jupyter-with-display-buffer'.")
     (unless buffer
       (setq buffer (get-buffer-create bname))
       (with-current-buffer buffer
+        ;; For buffers such as the jupyter REPL, showing trailing whitespaces
+        ;; may be a nuisance (as evidenced by the Python banner).
+        (setq-local show-trailing-whitespace nil)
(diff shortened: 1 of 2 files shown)


dzop:
Thanks.

Issue #87, opened 2019-04-10
R support


jackkamm:
Thanks for this amazing package.

I would like to use it with R. My initial attempts failed with the following message when I tried to run `jupyter-run-repl` or executing a `jupyter-R` src block:
```
ess--fl-keywords-values: ‘ess-font-lock-keywords’ must be a symbol of a custom variable
```

This was the stacktrace:
```
Debugger entered--Lisp error: (error "‘ess-font-lock-keywords’ must be a symbol of a custom variable")
  signal(error ("‘ess-font-lock-keywords’ must be a symbol of a custom variable"))
  error("`ess-font-lock-keywords' must be a symbol of a custom variable")
  ess--fl-keywords-values()
  ess--extract-fl-keywords()
  ess-build-font-lock-keywords()
  font-lock-eval-keywords(ess-build-font-lock-keywords)
  font-lock-set-defaults()
  font-lock-mode-internal(t)
  font-lock-default-function(t)
  font-lock-mode()
  jupyter-repl-initialize-fontification()
  jupyter-repl-mode()
  jupyter-repl--new-repl(#<jupyter-org-client jupyter-org-client> nil)
  jupyter-run-repl("R" nil nil jupyter-org-client)
  org-babel-jupyter-initiate-session-by-key("bar" ((:colname-names) (:rowname-names) (:result-params "replace") (:result-type . value) (:results . "replace") (:exports . "code") (:cache . "no") (:noweb . "no") (:hlines . "no") (:tangle . "no") (:async . "no") (:kernel . "R") (:session . "bar")))
  org-babel-jupyter-initiate-session("bar" ((:colname-names) (:rowname-names) (:result-params "replace") (:result-type . value) (:results . "replace") (:exports . "code") (:cache . "no") (:noweb . "no") (:hlines . "no") (:tangle . "no") (:async . "no") (:kernel . "R") (:session . "bar")))
  org-babel-execute:jupyter-R("list.files()" ((:colname-names) (:rowname-names) (:result-params "replace") (:result-type . value) (:results . "replace") (:exports . "code") (:cache . "no") (:noweb . "no") (:hlines . "no") (:tangle . "no") (:async . "no") (:kernel . "R") (:session . "bar")))
[...]

PR #89, opened 2019-04-11
R support

jackkamm:
This is a PR to add R support to emacs-jupyter.

Continuing the discussion from https://github.com/dzop/emacs-jupyter/issues/87:
1) I don't know of any non-ESS R-modes that are actively used/developed. Also, my post originally used `setq` on `ess-font-lock-keywords` but I've since changed it to `setq-local` to ensure it's only local in the REPL -- does this address your concerns about this?
2) In terms of ugly output -- the R kernel returns `text/html` versions of lists and tables, which `emacs-jupyter` prints by default, for example, here is how a table gets printed
```
#+begin_src jupyter-R :kernel R :session R
  data.frame(column1=1:5, column2=6:10, columne3=11:15)
#+end_src

#+RESULTS:
#+begin_export html
<table>
<thead><tr><th scope=col>column1</th><th scope=col>column2</th><th scope=col>columne3</th></tr></thead>
<tbody>
	<tr><td>1 </td><td> 6</td><td>11</td></tr>
	<tr><td>2 </td><td> 7</td><td>12</td></tr>
	<tr><td>3 </td><td> 8</td><td>13</td></tr>
	<tr><td>4 </td><td> 9</td><td>14</td></tr>
	<tr><td>5 </td><td>10</td><td>15</td></tr>
</tbody>
</table>
#+end_export
```
Setting `:display plain` prints the table in a more readable form, but it would be best to output as an `org-element`.

### DIFF
--- a/jupyter-R.el
+++ b/jupyter-R.el
@@ -1,0 +1,111 @@
+;;; jupyter-R.el --- Jupyter support for R -*- lexical-binding: t -*-
+
+;; Copyright (C) 2019 Nathaniel Nicandro
+
+;; Author: Jack Kamm <jackkamm@gmail.com>, Nathaniel Nicandro <nathanielnicandro@gmail.com>
+
+;; This program is free software; you can redistribute it and/or
+;; modify it under the terms of the GNU General Public License as
+;; published by the Free Software Foundation; either version 2, or (at
+;; your option) any later version.
+
+;; This program is distributed in the hope that it will be useful, but
+;; WITHOUT ANY WARRANTY; without even the implied warranty of
+;; MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
+;; General Public License for more details.
+
+;; You should have received a copy of the GNU General Public License
+;; along with GNU Emacs; see the file COPYING.  If not, write to the
+;; Free Software Foundation, Inc., 59 Temple Place - Suite 330,
+;; Boston, MA 02111-1307, USA.
+
+;;; Commentary:
+
+;; Support methods for integration with R.
+
+;;; Code:
+
+(require 'jupyter-repl)
+
+(cl-defmethod jupyter-repl-initialize-fontification (&context (jupyter-lang R))
+  (when (featurep 'ess)
+      (setq-local ess-font-lock-keywords 'ess-R-font-lock-keywords))
+  (cl-call-next-method))
+
+(cl-defmethod jupyter-org-result ((_mime (eql :text/html)) params data
+                                  &context (jupyter-lang R)
+                                  &optional metadata)
+  "Parse and convert DATA of type 'text/html' into a prettier form.
+If _METADATA has ':isolated' tag (e.g. DT::datatable()), save it to a file
+and open in a browser. Otherwise, if libxml is available and DATA
+is of type <table>, <ol>, <ul>, or <dl>, try to render it nicely."
+  (if (plist-get metadata :isolated)
+      ;; handle html with "isolated" property
+      (let ((file (or (alist-get :file params)
+                      (jupyter-org-image-file-name data ".html"))))
+        (with-temp-file file
+          (insert data))
+        (browse-url-of-file file)
+        (jupyter-org-file-link file))
+    ;; handle html without "isolated" property
+    (if (functionp 'libxml-parse-html-region)
+        ;; parse with libxml
+        (with-temp-buffer
+          (insert data)
+          (let* ((parsed-html (libxml-parse-html-region
+                               (point-min) (point-max)))
+                 (inner-element
+                  (jupyter-org-result--R-html-inner-element parsed-html))
+                 (inner-element-type (car inner-element)))
+            (cl-case (car inner-element)
+              (table
+               (jupyter-org-result--R-html-table parsed-html))
+              ((ol dl ul)
+               (jupyter-org-result--R-html-list parsed-html))
+              (t (cl-call-next-method)))))
+      ;; libxml not available, call next method
+      (cl-call-next-method))))
+
+(defun jupyter-org-result--R-html-table (parsed-html)
+  "Convert PARSED-HTML into an org-element table."
+  (with-temp-buffer
+    (shr-insert-document parsed-html)
+    (org-table-convert-region (point-min) (point-max) nil)
+    (org-table-insert-hline)
+    (let* ((tbl (org-table-to-lisp))
+           (head (jupyter-org-result--R-html-table-fix-header (car tbl)))
+           (body (cdr tbl)))
+      (jupyter-org-scalar `(,head . ,body)))))
+
+(defun jupyter-org-result--R-html-table-fix-header (header)
+  "Shifts column names to the right if last element of HEADER is empty string.
+This fixes the header column alignment when the table has row names."
+  (if (equal "" (car (last header)))
+      `("" . ,(butlast header))
+    header))
+
+(defun jupyter-org-result--R-html-list (parsed-html)
+  "Renders list-like PARSED-HTML.
+Since these can have arbitrary nested structure, we don't try to parse it into
+an org-element; instead we just render the html.  Note this is still a bit
+uglier than ':display plain', which should be preferred for these outputs."
+  (with-temp-buffer
+    (shr-insert-document parsed-html)
+    (buffer-string)))
+
+(defun jupyter-org-result--R-html-inner-element (parsed-html)
+  "If PARSED-HTML consists of a body with 1 element, return that element.
+Otherwise, return nil."
+  (let ((html-subnodes (cdr (cdr parsed-html))))
+    ;; check there is 1 html-subnode and it is the body
+    (when (eq 1 (length html-subnodes))
+      (let ((body (car html-subnodes)))
+        (when (eq (car body) 'body)
+          ;; check the body has 1 subnode and return it
+          (let ((body-subnodes (cdr (cdr body))))
+            (when (eq 1 (length body-subnodes))
+              (car body-subnodes))))))))
+
+(provide 'jupyter-R)
+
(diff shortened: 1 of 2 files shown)


jackkamm:
I've pushed an initial attempt to support tables in `jupyter-R.el`. The example before now looks better:
```
#+begin_src jupyter-R :kernel R :session R
  data.frame(column1=1:5, column2=6:10, columne3=11:15)
#+end_src

#+RESULTS:
| column1 | column2 | columne3 |
|---------+---------+----------|
|       1 |       6 |       11 |
|       2 |       7 |       12 |
|       3 |       8 |       13 |
|       4 |       9 |       14 |
|       5 |      10 |       15 |
```
I implemented it by overriding `jupyter-org-result` when `&context (jupyter-lang R)`. My solution uses libxml so I should also check that `(functionp 'libxml-parse-html-region)`, but I'm not sure how to add multiple checks into the `&context` block (this `cl-lib` stuff is new to me).

TODO: I'd also like to add support for lists and HTML widgets.

dzop:
> does this address your reservation?

If ESS is the only package used for R, its OK. I was thinking more of adding a check like

```emacs-lisp
(when <ESS is being used for R>
  (setq-local ...))
```

> I'm not sure how to add multiple checks into the `&context` block

The `&context` block only accepts a single expression so you can't combine the `jupyter-lang` specializer with something else, you will have to use something like the following

```emacs-lisp
&context ((and (eq (jupyter-kernel-language jupyter-current-client) 'R)
               (functionp 'libxml-parse-html-region))
          (eql t))
```

to check for `R` and `libxml-parse-html-region` in `&context`.  This is a little verbose, maybe it would be better just to use `(cl-call-next-method)` in the method body and fall back to a less specialized method, i.e. the default one, when `libxml-parse-html-region` is unavailable.

jupyter-R.el
```diff
@@ -11,12 +11,17 @@
         (t (cl-call-next-method))))
 
 (defun jupyter-org-result--R-html (data)
+  "Parse and convert DATA of type 'text/html' into a prettier form.
+If DATA is not a <table>, <ol>, or <dl>, returns nil."
```
dzop: We should also include `<ul>` here, right?

jupyter-R.el
```diff
@@ -11,12 +11,17 @@
...
   (with-temp-buffer
     (insert data)
     (let* ((parsed-html (libxml-parse-html-region (point-min) (point-max)))
-           (inner-element (jupyter-org-result--R-html-inner-element parsed-html)))
-      (cond ((eq (car inner-element) 'table)
-             (jupyter-org-result--R-html-table parsed-html))))))
+           (inner-element (jupyter-org-result--R-html-inner-element parsed-html))
+           (inner-element-type (car inner-element)))
+      (cond ((eq inner-element-type 'table)
```
dzop: We can also do

```emacs-lisp
(cl-case (car inner-element)
  (table
   ...)
  ((ol dl ul)
   ...))
```

its a little more compact.

jupyter-R.el
```diff
@@ -36,6 +41,15 @@ This fixes the header column alignment when the table has row names."
       `("" . ,(butlast header))
     header))
 
+(defun jupyter-org-result--R-html-list (parsed-html)
+  "Renders list-like html outputs.
+Since these can have arbitrary nested structure, we don't try to parse it into
```
dzop: Can we not recursively parse out the list structure here? Or are you saying that there can be stuff in the lists that org syntax doesn't allow as being part of a list?

jupyter-R.el
jackkamm: I so far haven't seen the R kernel output `<ul>`, but I've added it in case.

jupyter-R.el
jackkamm: Nice! I've done this as you suggested.

jupyter-R.el
jackkamm: I haven't extensively tested all the possibilities. But, for example, it's possible for it to return a list of tables, or list of lists of lists, etc. I'm not sure how well org can render these sorts of structures. It seemed simpler to just render the html with `shr.el` for now.

jackkamm:
Sorry @dzop , I did a squash & force push before I saw your code review. Hope it's not too confusing. I incorporated your suggestions in a subsequent commit.

This branch has reached a point that I'm more or less satisfied with, at least for an initial version. Here are some notes.

The R kernel has a few quirks that make it's behavior different. Most noticeably, tables and lists get returned in HTML form. While the tables are easy enough to convert to org, the lists can be fairly heterogeneous, and I'm not sure the correct way to convert them to org-elements. For example, we could have a list of lists of tables and lists. My solution for now is to just use `shr.el` to render lists. While this looks better than raw html, IMO `:display plain` looks much better. Ideally, we could prefer `:display plain` when rendering a list, and `:display html` when rendering a table, but I think this would add too much complexity, and I think it's fine to let the user specify the `:display` type as needed.

Another area where the R kernel seems to differ is with HTML widgets. I tested this out on the package [DT](https://rstudio.github.io/DT/), which creates a nice javascript table that can be sorted, filtered, etc in a browser, and is an important part of my data analysis workflow. Unlike other kernels, the R kernel doesn't return an `application/vnd.jupyter.widget-view+json` -- instead, it returns an HTML file, with an `isolated` metadata tag. My solution was to paste this html into a file, return a link to the file, and open it with `browse-url`.

A final issue is that `jupyter-eval-string` (and related functions) only partially work. They send code to the REPL fine, but no output is returned from the REPL (unlike Python, which displays the output in a nice message or pop-up buffer). I'm not 100% sure, but I think what's happening is that the R kernel doesn't send an `execute_result` message, which `jupyter-eval-string` uses to capture the output.

While still having some rough edges, I think `jupyter-R.el` is good enough that I can use it to do my work, so if it's alright with you I would favor merging it in and fixing the outstanding issues in future issues/PRs. But if you want me to spend a bit more time cleaning up before merging, that's fine with me too. Since you don't use R, I'd be willing to volunteer on any future issues/PRs involving `jupyter-R.el`.

jupyter-R.el
dzop: OK, yeah I don't think org can handle all of that.

dzop:
About `jupyter-eval-string`, it is probably because only a `:display-data` message is being sent back. It looks like I forgot to add a callback function for `:display-data` messages.

Can you set `jupyter--debug` to `t`, evaluate code that doesn't show any results when it should, and look to see if there are `:display-data` messages in the `*Messages*` buffer to verify.

jackkamm:
Yep, I see a `display-data` -- here's the full output from evaluating the list `1:10`:
```
SENDING: :execute-request (:code 1:10 :silent :json-false :store_history :json-false :user_expressions #s(hash-table size 1 test eql rehash-size 1.5 rehash-threshold 0.8125 data ()) :allow_stdin t :stop_on_error :json-false)
SENT: (:shell 5615fb34-99d5-4f0e-bf07-dc52d9986577)
MESSAGE: (:iopub :status (:execution_state busy))
MESSAGE: (:iopub :execute-input (:code 1:10 :execution_count 3))
DROPPING-REQ: 5d05aa7b-6098-4da3-b45f-f020df48802d
MESSAGE: (:iopub :display-data (:data (:text/html <ol class=list-inline>
	<li>1</li>
	<li>2</li>
	<li>3</li>
	<li>4</li>
	<li>5</li>
	<li>6</li>
	<li>7</li>
	<li>8</li>
	<li>9</li>
	<li>10</li>
</ol>
 :text/markdown 1. 1
2. 2
3. 3
4. 4
5. 5
6. 6
7. 7
8. 8
9. 9
10. 10


 :text/latex \begin{enumerate*}
\item 1
\item 2
\item 3
\item 4
\item 5
\item 6
\item 7
\item 8
\item 9
\item 10
\end{enumerate*}
 :text/plain  [1]  1  2  3  4  5  6  7  8  9 10) :metadata nil))
MESSAGE: (:shell :execute-reply (:status ok :execution_count 3 :payload [] :user_expressions nil))
MESSAGE: (:iopub :status (:execution_state idle))
jupyter: eval done
```

jupyter-R.el
```diff
@@ -0,0 +1,111 @@
...
+(require 'jupyter-repl)
+
+(cl-defmethod jupyter-repl-initialize-fontification (&context (jupyter-lang R))
+  (if (featurep 'ess)
+      (setq-local ess-font-lock-keywords 'ess-R-font-lock-keywords))
+  (cl-call-next-method))
+
+(cl-defmethod jupyter-org-result ((_mime (eql :text/html)) _params data
+                                  &context (jupyter-lang R)
+                                  &optional _metadata)
```
dzop: You should change `_metadata` and `_params` to just `metadata` and `params` since you use the arguments in the method body. The underscore is used to indicate that an argument is **not** used inside the method and so that the byte compiler does not complain about unused variables.

jupyter-R.el
```diff
@@ -0,0 +1,111 @@
...
+;;; Commentary:
+
+;; Support methods for integration with R.
+
+;;; Code:
+
+(require 'jupyter-repl)
+
+(cl-defmethod jupyter-repl-initialize-fontification (&context (jupyter-lang R))
+  (if (featurep 'ess)
```
dzop: Could you use `when` here.

dzop:
LGTM. Will merge after you get a chance to address my new comments. Also the issue with `jupyter-eval-string` should be resolved on the latest master.

jupyter-R.el
jackkamm: Done

jupyter-R.el
jackkamm: Fixed

jackkamm:
I've added your suggestions, rebased, squashed and force pushed. I checked and `jupyter-eval-string` works now with the R kernel. Thanks! Loving this package.

dzop:
Thanks!

Issue #88, opened 2019-04-10
Support pandas DataFrames


qleguennec:
I would like to output pandas `DataFrames` objects in org-tables format
See:

```
#+BEGIN_SRC jupyter-python :session py :async yes :display text/plain
  data = %sql select * from actor where first_name like 'E%'
  dataframe = data.DataFrame()
  dataframe
#+END_SRC

#+RESULTS:
:RESULTS:
:  * postgresql://admin:***@localhost/dvdrental
: 7 rows affected.
:    actor_id first_name  last_name             last_update
: 0         3         Ed      Chase 2013-05-26 14:47:57.620
: 1        22      Elvis       Marx 2013-05-26 14:47:57.620
: 2        93      Ellen    Presley 2013-05-26 14:47:57.620
: 3       136         Ed  Mansfield 2013-05-26 14:47:57.620
: 4       139       Ewan    Gooding 2013-05-26 14:47:57.620
: 5       148      Emily        Dee 2013-05-26 14:47:57.620
: 6       179         Ed    Guiness 2013-05-26 14:47:57.620
:END:
```

dcherian:
See https://bitbucket.org/astanin/python-tabulate .

I have trouble getting it to output in a format that I can export easily though. The output keeps getting wrapped in an `example` block.

dzop:
Using `tabulate` is the way to go here since `emacs-jupyter` only has access to the mimetype information of the results, e.g. if it is html  or png. If you look at the last tip in https://github.com/gregsexton/ob-ipython/blob/7147455230841744fb5b95dcbe03320313a77124/README.org#tips-and-tricks, you can get `org-mode` table output for dataframes.

qleguennec:
@dzop Thanks, that works.

iwrotesomecode:
> @dzop Thanks, that works.

@qleguennec 
I dropped that linked code in my ipython startup files `~/.ipython/profile_default/startup/ ` and still get datatables in example blocks not formatted as org-tables. `Tabulate` is verified installed. Did you have to do anything else to get this working?

PR #91, opened 2019-04-13
Fix R parsing HTML table with spaces, using pandoc

jackkamm:
Sorry to be revisiting this so soon -- I found a bug already in `jupyter-R.el` :-/

If a table has spaces, it gets parsed incorrectly, like so:
```
#+begin_src jupyter-R :session R
  data.frame(x=c("a", "b c", "d"), y=1:3)
#+end_src

#+RESULTS:
|   | x | y |
|---+---+---|
| a | 1 |   |
| b | c | 2 |
| d | 3 |   |
```
The second entry of the "x" column should be "b c" but it gets split into an extra column, and the rest of the table is mangled as a consequence.

I decided to try an alternate approach. If `pandoc.el` is available, use it to convert HTML output to org-mode. Otherwise, if libxml is available, use shr to render it -- don't try to do any HTML parsing ourselves. This simplifies the code a bit, hopefully making it more maintainable.

Here's some examples. Table output when pandoc is available:
```
#+begin_src jupyter-R :session R
  data.frame(x=c("a", "b c", "d"), y=1:3)
#+end_src

#+RESULTS:
:RESULTS:
| x     | y   |
|-------+-----|
| a     | 1   |
| b c   | 2   |
| d     | 3   |
:END:
```
Libxml table output:
```
#+begin_src jupyter-R :session R
  data.frame(x=c("a", "b c", "d"), y=1:3)
#+end_src

#+RESULTS:
:   x  y    
:   a   1     
:   b c  2     
:   d   3     
```
Pandoc output of structured list:
```
#+begin_src jupyter-R :session R :display html
  foo = data.frame(a=letters[1:3], b=1:3)
  list(foo, 1:3, c=foo)
#+end_src

#+RESULTS:
:RESULTS:
- [[1]] :: | a   | b   |
  |-----+-----|
  | a   | 1   |
  | b   | 2   |
  | c   | 3   |

- [[2]] :: 

  1. 1
  2. 2
  3. 3

- $c :: | a   | b   |
  |-----+-----|
  | a   | 1   |
  | b   | 2   |
  | c   | 3   |
:END:
```
The nested list rendering is not perfect -- the inner table headers are shifted, and [[1]] and [[2]] get rendered as links. However it's still readable and not a huge deal IMO -- can always use `:display plain` for these sorts of edge cases.

An alternative which could be cool (if you're open to it) would be to support `pandoc.el` more generally, perhaps through a `:results pandoc` header property. Then we could remove the special handling of R html results. It might also be nice for pandas dataframes (https://github.com/dzop/emacs-jupyter/issues/88) -- that way people don't have to mess with their ipython startup files.

### DIFF
--- a/jupyter-org-client.el
+++ b/jupyter-org-client.el
@@ -845,6 +845,36 @@ the width or height of the image."
         metadata
       (jupyter-org-image-link file width height))))
 
+(defun jupyter-pandoc-convert (from to from-string &optional callback)
+  "Use pandoc to convert a string in FROM format to TO format.
+Starts a process and converts FROM-STRING, assumed to be in FROM
+format, to a string in TO format and returns the converted
+string.
+
+If CALLBACK is specified, return the process object. When the
+process exits, call CALLBACK with zero arguments and with the
+buffer containing the converted string current."
+  (cl-assert (executable-find "pandoc"))
+  (let* ((process-connection-type nil)
+         (proc (start-process
+                "jupyter-pandoc"
+                (generate-new-buffer " *jupyter-pandoc*")
+                "pandoc" "-f" from "-t" to "--")))
+    (set-process-sentinel
+     proc (lambda (proc _)
+            (when (memq (process-status proc) '(exit signal))
+              (with-current-buffer (process-buffer proc)
+                (funcall callback)
+                (kill-buffer (process-buffer proc))))))
+    (process-send-string proc from-string)
+    (process-send-eof proc)
+    (if callback proc
+      (let ((to-string ""))
+        (setq callback (lambda () (setq to-string (buffer-string))))
+        (while (zerop (length to-string))
+          (accept-process-output nil 1))
+        to-string))))
+
 (cl-defgeneric jupyter-org-result (_mime _params _data &optional _metadata)
   "Return an `org-element' representing a result.
 Either a string or an `org-element' is a valid return value of
@@ -991,9 +1021,18 @@ parsed, wrap DATA in a minipage environment and return it."
       (jupyter-org--parse-latex-element data)
     (jupyter-org-export-block "latex" data)))
 
-(cl-defmethod jupyter-org-result ((_mime (eql :text/html)) _params data
+(cl-defmethod jupyter-org-result ((_mime (eql :text/html)) params data
                                   &optional _metadata)
-  (jupyter-org-export-block "html" data))
+  (if (member "pandoc" (alist-get :result-params params))
+      (jupyter-org-scalar
+       (with-temp-buffer
+         (insert (jupyter-pandoc-convert "html" "org" data))
+         (goto-char (point-min))
+         (if (and (org-at-table-p)
+                  (eq (org-table-end) (point-max)))
+             (org-table-to-lisp)
+           (buffer-string))))
+    (jupyter-org-export-block "html" data)))
 
 ;; NOTE: The order of :around methods is that the more specialized wraps the
 ;; more general, this makes sense since it is how the primary methods work as
--- a/jupyter-org-client.el
+++ b/jupyter-org-client.el
@@ -1028,6 +1028,7 @@ parsed, wrap DATA in a minipage environment and return it."
        (with-temp-buffer
          (insert (jupyter-pandoc-convert "html" "org" data))
          (goto-char (point-min))
+         (delete-trailing-whitespace)
          (if (and (org-at-table-p)
                   (eq (org-table-end) (point-max)))
              (org-table-to-lisp)
--- a/jupyter-R.el
+++ b/jupyter-R.el
@@ -35,29 +35,16 @@
 (cl-defmethod jupyter-org-result ((_mime (eql :text/html)) params data
                                   &context (jupyter-lang R)
                                   &optional metadata)
-  "Parse and convert DATA of type 'text/html' into a prettier form.
-If METADATA has ':isolated' tag (e.g. DT::datatable()), save it to a file
-and open in a browser. Otherwise, try to use pandoc or libxml to convert
-DATA to a prettier form."
-  (cond
-   ((plist-get metadata :isolated)
+  "If html DATA is an iframe, save it to a separate file and open in browser.
+Otherwise, parse it as normal."
+  (if (plist-get metadata :isolated)
       (let ((file (or (alist-get :file params)
                       (jupyter-org-image-file-name data ".html"))))
         (with-temp-file file
           (insert data))
         (browse-url-of-file file)
-        (jupyter-org-file-link file)))
-   ((functionp 'pandoc-convert-stdio)
-    (jupyter-org-raw-string (pandoc-convert-stdio data "html" "org")))
-   ((functionp 'libxml-parse-html-region)
-    (with-temp-buffer
-      (insert data)
-      (let ((parsed-html (libxml-parse-html-region
-                          (point-min) (point-max))))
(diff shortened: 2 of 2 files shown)


dzop:
I think it makes sense to rely on `pandoc` for processing these kinds of cases. The `nbconvert` program of Jupyter already uses pandoc so its most likely available for users. There are most likely other instances where a kernel emits HTML that could be transformed into org syntax. I've also been experimenting with converting notebooks to org-mode documents using pandoc (https://github.com/dzop/emacs-jupyter/tree/feature/notebook-conversion) so I would have probably ended up doing something similar. I don't think we need to rely on `pandoc.el` though, we can just roll our own function:

```emacs-lisp
(defun jupyter-pandoc-convert (from to from-string &optional callback)
  "Use pandoc to convert a string in FROM format to TO format.
Starts a process and converts FROM-STRING, assumed to be in FROM
format, to a string in TO format and returns the converted
string.

If CALLBACK is specified, return the process object. When the
process exits, call CALLBACK with zero arguments and with the
buffer containing the converted string current."
  (cl-assert (executable-find "pandoc"))
  (let* ((process-connection-type nil)
         (proc (start-process
                "jupyter-pandoc"
                (generate-new-buffer " *jupyter-pandoc*")
                "pandoc" "-f" from "-t" to "--")))
    (set-process-sentinel
     proc (lambda (proc _)
            (when (memq (process-status proc) '(exit signal))
              (with-current-buffer (process-buffer proc)
                (funcall callback)
                (kill-buffer (process-buffer proc))))))
    (process-send-string proc from-string)
    (process-send-eof proc)
    (if callback proc
      (let ((to-string ""))
        (setq callback (lambda () (setq to-string (buffer-string))))
        (while (zerop (length to-string))
          (accept-process-output nil 1))
        to-string))))
```

Also, I forgot to mention this last time, but could you ensure that your changes byte compile cleanly by addressing the warnings you get when you run `byte-compile-file` on `jupyter-R.el`.

jackkamm:
I've run `byte-compile-file` and addressed the warnings, sorry about that.

Glad you're supportive of using pandoc to render HTML more generally. Therefore, I think it makes sense to remove the special pandoc/shr handling from `jupyter-R.el` entirely, which I've done.

I added general pandoc rendering for HTML in a standalone commit. There's a couple ways to go about this, and I'm not sure I picked the way you'd like, so we can reset and pursue in another branch if you prefer that.

Specifically, I added a `:results pandoc` option for HTML output, which calls out to your `jupyter-pandoc-convert`, plus a little code I added to render tables specially. Below are a couple alternative possibilities...what do you think?
1. Get rid of `:results pandoc`, and always render HTML with pandoc by default. Maybe add a `:results raw` option to render raw HTML, or do some special checking to make sure the HTML is something we'd actually want to convert (e.g. list, table, etc.)
2. At the other extreme, keep the `:results pandoc`, and also provide this option to latex and markdown outputs as well. For example, I did notice slight differences between the pandoc conversions of Markdown and HTML outputs on nested R lists. However, this is an edge case and the differences were minor.

jupyter-org-client.el
```diff
@@ -991,9 +1021,18 @@ parsed, wrap DATA in a minipage environment and return it."
       (jupyter-org--parse-latex-element data)
     (jupyter-org-export-block "latex" data)))
 
-(cl-defmethod jupyter-org-result ((_mime (eql :text/html)) _params data
+(cl-defmethod jupyter-org-result ((_mime (eql :text/html)) params data
                                   &optional _metadata)
-  (jupyter-org-export-block "html" data))
+  (if (member "pandoc" (alist-get :result-params params))
+      (jupyter-org-scalar
```
jackkamm: On further testing, `jupyter-org-scalar` seems to append an extra line if `data` is a short string (but not if it's a long string wrapped in a quote block). It seems to be necessary when data is a table though. I'm slightly confused when I should be calling `jupyter-org-scalar`...

jupyter-org-client.el
```diff
@@ -845,6 +845,36 @@ the width or height of the image."
         metadata
       (jupyter-org-image-link file width height))))
 
+(defun jupyter-pandoc-convert (from to from-string &optional callback)
```
dzop: Could you move this to `jupyter-mime.el` since it could be useful in other contexts in the future. You can place it at the end of the file with a new comment heading:

```emacs-lisp
;;; Pandoc
```

jupyter-org-client.el
dzop: Yes `jupyter-org-scalar` should only be passed an unformatted string or table. It is not intended to handle `org` formatted strings. If you want to handle tables specially you should wrap the `org-tabel-to-lisp` call with `jupyter-org-scalar` and the `buffer-string` call with `jupyter-org-raw-string`.

Couldn't you avoid doing all of the table processing by just doing something like

```emacs-lisp
(jupyter-org-raw-string (jupyter-pandoc-convert "html" "org" data))
```

dzop:
I would prefer your option (1) to option (2), but I don't think we should do any checking of the HTML since there is most likely some quirks with how the various kernel languages emit HTML that we would have to take into account. We can revisit this if needed in the future.

I think `:results raw` should have the opposite behavior, i.e. specifying `:results raw` should use `pandoc` to convert the HTML to org syntax since that is closer to the intended meaning of `:results raw` (see `org-babel-insert-result`). I would like to keep the meaning of the header arguments as consistent as possible with how `org-mode` itself uses them to avoid any confusion. At the very least I would prefer not to add any new options to `:results`.

Another alternative would be to:

3. Provide a new header argument, `:pandoc t`, in combination with a customizable variable `jupyter-org-pandoc-convertable` (or a better name) which is a list of mime types that are convertable to org syntax using pandoc. Specifying `:pandoc t` on a code block means that if the kernel emits a mime type in `jupyter-org-pandoc-convertable`, then the associated data is converted into org syntax using pandoc. 

(3) seems like a better approach than (1) or (2) as it leaves it up to the user to specify which mime types are convertable using pandoc. What do you think?

Also I think it would be a good idea to close this pull request and open up two new ones. One with the pandoc feature and another with the changes to `jupyter-R.el`.

jupyter-org-client.el
jackkamm: Reason I was using table parsing and `jupyter-org-scalar`, is that `jupyter-org-raw-string` inserts the table in a `:RESULTS:` drawer, and I was trying to replicate the behavior of `jupyter-python` and `ob-python` which doesn't have a drawer. However, I think you're right, it's better to be simple, so I'll switch to `jupyter-org-raw-string` without the table parsing.

Longer term, it might be nice to have finer-grained control over whether the table should go in a drawer. For example, I prefer a drawer for big tables but not small ones. But this is not a priority for me, and should probably be a separate issue.

jackkamm:
(3) sounds great to me. Closing this to split into 2 separate PRs which I'll submit shortly.

jupyter-org-client.el
dzop: I think you can get around wrapping the table with a drawer by also putting an `org-table` text property on the 1st character of the table string. The result won't be wrapped in a drawer if it can be removed by `org-babel` and determining if a result needs to be wrapped is mainly handled by the function `jupyter-org-babel-result-p`.

It would probably be good to create a new function to add this text property similar to what `jupyter-org-raw-string` and `jupyter-org-raw-string-p` do but for the `org-table` property. When doing this we will have to replace all instances of the `org-table` property with the corresponding functions.

PR #97, opened 2019-04-24
Add pandoc conversion of rich outputs

jackkamm:
See also: https://github.com/dzop/emacs-jupyter/pull/91

As we discussed tables are currently output inside `:RESULTS:` drawers. I haven't implemented your suggestion to add an `org-table` text property, or the special parsing needed to determine when the property should be added. I've left that for a future issue/PR -- but let me know if you prefer I take care of it now, in which case I'll look into it.

### DIFF
--- a/jupyter-org-client.el
+++ b/jupyter-org-client.el
@@ -66,6 +66,12 @@ automatically be shown if this is non-nil."
   :group 'ob-jupyter
   :type 'boolean)
 
+(defcustom jupyter-org-pandoc-convertable
+  '("html" "markdown" "latex")
+  "Export blocks to convert to org-mode when ':pandoc t' header is set."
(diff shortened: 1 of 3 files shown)


jupyter-org-client.el
```diff
@@ -648,6 +654,16 @@ inserted without modification as the result of a code block."
   "Return a comment `org-element' with VALUE."
   (list 'comment (list :value value)))
 
+(defun jupyter-org-export-or-pandoc (type value params)
```
dzop: Can we rename this to `jupyter-org-export-block-or-pandoc`? It makes it slightly more obvious what the function is doing.

jackkamm: Fixed.

PR #107, opened 2019-05-06
Fix two org-src-lang-modes calc issues

xuning97:
For assoc lang org-src-lang-modes may return symbol sh, (inter 'sh) will
return an error.

For some languages like C++, the corresponding language mode shall be c++

### DIFF
--- a/ob-jupyter.el
+++ b/ob-jupyter.el
@@ -435,9 +435,9 @@ Optional argument REFRESH has the same meaning as in
                         (cdr (assoc lang org-babel-tangle-lang-exts)))))
    (add-to-list 'org-src-lang-modes
                 (cons (concat "jupyter-" lang)
-                      (intern (or (cdr (assoc lang org-src-lang-modes))
-                                  (replace-regexp-in-string
-                                   "[0-9]*" "" lang)))))))
+                      (or (cdr (assoc lang org-src-lang-modes))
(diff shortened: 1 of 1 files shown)


xuning97:
@dzop May you kindly check if my understanding is correct

dzop:
Yes your right, for C++ `org-src-lang-modes` has an entry like

```("jupyter-C++17" . C++)```

where the mode `C++` should be `c++`. And I don't know what that outer `intern` call is doing. Good catch.

How about simplifying your change to just remove the outer `intern` call:

```emacs-lisp
(or (cdr (assoc lang org-src-lang-modes))
    (downcase (replace-regexp-in-string
               "[0-9]*" "" lang)))
```

No need to call `intern` since `org-src-lang-modes` says that the mode can be either a string or symbol.

xuning97:
I’ll do that

dzop:
Yes, the default values for the modes are all symbols, but the documentation says they can be either strings or symbols and the function which uses `org-src-lang-modes`, `org-src--get-lang-mode`, handles both cases. It actually just uses the mode as is if it isn't a symbol.

But if you prefer it to be a symbol, I'm fine with that. I don't think we need to explicitly check if the value we get from `org-src-lang-modes` is  a symbol though.

xuning97:
As long as it's documented :-)

I've updated the code.

PR #108, opened 2019-05-07
Add prev/next-matching REPL-input-history funcs

poppyschmo:
This attempts to mimic the behavior offered by `comint-previous-matching-input-from-input` and `previous-complete-history-element` and their fwd/next counterparts and helpers functions.

Users of GNU Readline's `history-search-backward` (and Vim's `c_<Up>`) may also find it familiar. It's also part of the DWIM up/down/prev/next navigation of Terminal IPython.

Update: tidy up a bit, make numeric arg optional, confirm some facts about rings, rebase atop latest master


jupyter-repl.el
```diff
@@ -962,6 +962,69 @@ buffer to display TEXT."
...
+  (let* ((r jupyter-repl-history)
+         (n (ring-length r))
+         (direction (if (> arg 0) +1 -1))
+         (i (if (= direction -1) 0 -1)) ; adjust for initial increment
+         (failed (zerop arg))
+         ;;
+         code)
+    ;; Search ARG times
+    (while (not (or failed (zerop arg)))
+      (while (not (or (setq i (mod (+ i direction) n)
```
dzop: We don't need this `mod` here since `ring-ref` already does this. I believe we can just do

```emacs-lisp
code (ring-ref r (cl-incf i direction))
```

poppyschmo: Great, thanks. This also simplifies the post-helper figuring in the main command, since the sign of the returned index now matches that of the numeric arg.

poppyschmo:
Hmm, Travis doesn't seem very happy. Hope my test didn't introduce a race. A build of the same SHA  [passed](https://travis-ci.com/poppyschmo/emacs-jupyter/builds/111135613) minutes before.

dzop:
Yes the tests have been failing recently with a lot of errors like

`(error "Channel not stopped in ioloop subprocess (:shell)")`

And some others on Appveyor. Any tests about "ioloop". They are nothing to be concerned about. If you run the build a few times they should pass. I think it has something to do with starting/stopping lots of subprocesses too quickly, which is what the "ioloop" business is about, but haven't really figured it out yet.

poppyschmo:
Ah, gotcha. Thanks for explaining. Seems like you're nailing it down.

jupyter-repl.el
```diff
@@ -962,6 +962,67 @@ buffer to display TEXT."
...
+  (when (or (null n) (zerop n)) (setq n 1))
+  (let ((opoint (point))
+        (r jupyter-repl-history)
+        (code (jupyter-repl-cell-code))
+        (is-prev (> n 0))
+        (input-string (buffer-substring
+                       (jupyter-repl-cell-code-beginning-position) (point)))
+        found)
+    ;; Look past an initial duplicate
+    (if is-prev
```
dzop: We can simplify this to

```emacs-lisp
(when (equal code (ring-ref r (if is-prev 0 -1)))
  (cl-incf n (if is-prev 1 -1)))
```

We check `is-prev` twice, but it's less clutter. WDYT?

poppyschmo: Slick. I suppose there's no longer a real reason to initialize `code`, now, except perhaps for easier reading.

Side note: GitHub displayed a hint bubble saying something about my being able to integrate your fix on the fly, but I shooed it away in haste, unfortunately. Guess I'll see what the docs have to say.

jupyter-repl.el
```diff
@@ -962,6 +962,66 @@ buffer to display TEXT."
...
+  "Search input history for the input pending before point.
+On success, replace the current input with the matching code element
+while preserving and returning point. Ding on failure. If N is negative,
+find the Nth next match; if positive, the Nth previous. If N is zero or
+nil, pretend it's one."
+  ;; Adapted from: `comint-previous-matching-input-from-input' and friends
+  (interactive "p")
+  (when (or (null n) (zerop n)) (setq n 1))
+  (let ((opoint (point))
+        (r jupyter-repl-history)
```
dzop: Can we get rid of this `r` variable,  we can just use `jupyter-repl-history`.

jupyter-repl.el
```diff
@@ -962,6 +962,66 @@ buffer to display TEXT."
...
            beg end 'font-lock-face 'jupyter-repl-traceback)))))))
 
+(defun jupyter-repl-history--match-input (regexp arg)
+  "Return the index of the ARGth REGEXP match.
+Or nil, on failure. If ARG is positive, search backward from the most
+recent history element. If negative, search forward through items
+previously visited during this input session. If ARG is zero, do
+nothing."
+  ;; Adapted from `comint-previous-matching-input-string-position'
+  (let* ((r jupyter-repl-history)
```
dzop: See https://github.com/dzop/emacs-jupyter/pull/108/files#r284953826

jupyter-repl.el
poppyschmo: Removed the variable in both functions.

dzop:
I've pushed some changes to `master` that generalizes the `jupter-repl-history--next` and `jupyter-repl-history--previous` functions. I had to make the following change to your code

```diff
diff --git a/jupyter-repl.el b/jupyter-repl.el
index 7b89d941..525ed6bf 100644
--- a/jupyter-repl.el
+++ b/jupyter-repl.el
@@ -1006,9 +1006,7 @@ (defun jupyter-repl-history-previous-matching (&optional n)
                           (concat "^" (regexp-quote input-string)) n)))
         (user-error "No %s matching input" (if is-prev "earlier" "later"))
       (setq code (ring-ref r found))
-      (if is-prev
-          (jupyter-repl-history--previous found)
-        (jupyter-repl-history--next (- found)))
+      (jupyter-repl-history--rotate (- found))
       (jupyter-repl-replace-cell-code code)
       (goto-char opoint))))
```

Could you rebase on master and make the above change. After doing so, this patch should be good to go.

poppyschmo:
Rebased atop 005f2e30bfc9c0e9a27d28e7ea6f61c04e24a28b with the requested change. Thanks for including the diff.

PR #115, opened 2019-05-10
Handle `isolated` metadata property for HTML results of the R kernel

jackkamm:
I'd like to have the REPL handle html outputs with `<script>` tags. I've implemented this by using `browse-url-of-buffer` to open in an external browser when the `<script>` tag is detected.

My immediate motivation was to easily inspect R dataframes with the javascript DataTables library via  R's [DT](https://rstudio.github.io/DT/) package.

### DIFF
--- a/jupyter-client.el
+++ b/jupyter-client.el
@@ -977,9 +977,13 @@ to the above explanation."
         (setq had-result t)
         (jupyter-with-display-buffer "display" req
           (jupyter-insert data metadata)
-          (jupyter-display-current-buffer-reuse-window
-           :display-data nil (unless (jupyter-pop-up-frame-p :display-data)
-                               #'display-buffer-below-selected))))
+          (if (> (line-number-at-pos (point-max))
+                 jupyter-eval-short-result-max-lines)
+              (jupyter-display-current-buffer-reuse-window
+               :display-data nil (unless (jupyter-pop-up-frame-p :display-data)
+                                   #'display-buffer-below-selected))
+            (funcall jupyter-eval-short-result-display-function
+                     (buffer-string)))))
       :error
       (jupyter-message-lambda (traceback)
         ;; FIXME: Assumes the error in the
--- a/jupyter-client.el
+++ b/jupyter-client.el
@@ -977,9 +977,12 @@ to the above explanation."
         (setq had-result t)
         (jupyter-with-display-buffer "display" req
           (jupyter-insert data metadata)
-          (jupyter-display-current-buffer-reuse-window
-           :display-data nil (unless (jupyter-pop-up-frame-p :display-data)
-                               #'display-buffer-below-selected))))
+          ;; Don't pop-up the display when it's empty (e.g. jupyter-R
+          ;; will open some HTML results in an external browser)
+          (when (/= (point-min) (point-max))
+            (jupyter-display-current-buffer-reuse-window
+             :display-data nil (unless (jupyter-pop-up-frame-p :display-data)
+                                 #'display-buffer-below-selected)))))
       :error
       (jupyter-message-lambda (traceback)
         ;; FIXME: Assumes the error in the
--- a/jupyter-client.el
+++ b/jupyter-client.el
@@ -977,13 +977,9 @@ to the above explanation."
         (setq had-result t)
         (jupyter-with-display-buffer "display" req
           (jupyter-insert data metadata)
-          (if (> (line-number-at-pos (point-max))
-                 jupyter-eval-short-result-max-lines)
-              (jupyter-display-current-buffer-reuse-window
-               :display-data nil (unless (jupyter-pop-up-frame-p :display-data)
-                                   #'display-buffer-below-selected))
-            (funcall jupyter-eval-short-result-display-function
-                     (buffer-string)))))
+          (jupyter-display-current-buffer-reuse-window
+           :display-data nil (unless (jupyter-pop-up-frame-p :display-data)
+                               #'display-buffer-below-selected))))
       :error
       (jupyter-message-lambda (traceback)
         ;; FIXME: Assumes the error in the
--- a/jupyter-mime.el
+++ b/jupyter-mime.el
@@ -319,11 +319,10 @@ aligns on the current line."
     (browse-url-of-file file)
     ;; Give the external browser time to open the tmp file before deleting it
     ;; based on mm-display-external
-    (let ((file file))
-      (run-at-time
-       60.0 nil
-       (lambda ()
-         (ignore-errors (delete-file file)))))))
+    (run-at-time
+     60.0 nil
+     (lambda ()
+       (ignore-errors (delete-file file))))))
 
 (defun jupyter--delete-script-tags (beg end)
   (save-excursion
--- a/jupyter-mime.el
+++ b/jupyter-mime.el
@@ -313,6 +313,14 @@ aligns on the current line."
         (setf (image-property image :ascent) 50)
         (force-window-update)))))
 
+(defun jupyter-insert-html--maybe-browse-url (beg end)
+  (save-excursion
+    (save-restriction
+      (narrow-to-region beg end)
+      (goto-char beg)
+      (when (re-search-forward "<script[^>]*>" nil t)
+        (browse-url-of-buffer)))))
+
 (defun jupyter--delete-script-tags (beg end)
   (save-excursion
     (save-restriction
@@ -329,10 +337,9 @@ aligns on the current line."
   "Parse and insert the HTML string using `shr'."
   (jupyter-with-insertion-bounds
       beg end (insert html)
-    ;; TODO: We can't really do much about javascript so
-    ;; delete those regions instead of trying to parse
-    ;; them. Maybe just re-direct to a browser like with
-    ;; widgets?
+    ;; Maybe open the html in external browser (e.g. if it has javascript)
+    (jupyter-insert-html--maybe-browse-url beg end)
+    ;; Render non-javascript regions in the buffer
     ;; NOTE: Parsing takes a very long time when the text
     ;; is > ~500000 characters.
     (jupyter--delete-script-tags beg end)
--- a/jupyter-mime.el
+++ b/jupyter-mime.el
@@ -313,6 +313,18 @@ aligns on the current line."
         (setf (image-property image :ascent) 50)
         (force-window-update)))))
 
+(defun jupyter--insert-temp-file-browse-url (data)
+  (let ((file (make-temp-file "emacs-jupyter" nil ".html")))
+    (with-temp-file file (insert data))
+    (browse-url-of-file file)
+    ;; Give the external browser time to open the tmp file before deleting it
+    ;; based on mm-display-external
+    (lexical-let ((file file))
+      (run-at-time
+       60.0 nil
+       (lambda ()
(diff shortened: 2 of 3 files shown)


jupyter-mime.el
```diff
@@ -329,10 +337,9 @@ aligns on the current line."
   "Parse and insert the HTML string using `shr'."
   (jupyter-with-insertion-bounds
       beg end (insert html)
-    ;; TODO: We can't really do much about javascript so
-    ;; delete those regions instead of trying to parse
-    ;; them. Maybe just re-direct to a browser like with
-    ;; widgets?
+    ;; Maybe open the html in external browser (e.g. if it has javascript)
+    (jupyter-insert-html--maybe-browse-url beg end)
```
dzop: I think this should be dependent on kernel language. Although, it doesn't do it now, the Plots package from IJulia used to send a lot of embedded scripts to setup javascript plots before plotting and this javascript wasn't meant for viewing in a browser.

We can do something like

```emacs-lisp
(cl-defmethod jupyter-insert ((_mime (eql :text/html)) data
                              &context (jupyter-lang R)
                              &optional _metadata)
  (jupyter-with-insertion-bounds
      beg end (cl-call-next-method)
    (jupyter-insert-html--maybe-browse-url beg end)))

(cl-defmethod jupyter-insert :around ((_mime (eql :text/html)) data
                                      &optional _metadata)
  (jupyter-with-insertion-bounds
      beg end (cl-call-next-method)
    (jupyter--delete-script-tags beg end)))
```

What happens here is that the `:around` method gets called first which will call the `R` method, which in turn calls the base definition already present. So the `R` method gets a chance at checking from script tags before they get removed from the html in the `:around` method.

jackkamm:
OK, based on your review I've changed this to be language-specific. In particular I've pushed a commit to define a `jupyter-insert` method for `&context (jupyter-lang R)`. I had to remove the `&context` of the default `jupyter-insert`  for `text/html`, otherwise the R-version failed to override the default one.

I also thought it might be simpler to directly put the data in a temp file and `browse-url` on that, rather than inserting into the REPL and doing `browse-url` on the narrowed region, so I've switched to doing that. This avoids needing to write an `:around` version of `jupyter-insert` (I think).

One problem I just noticed, and haven't figured out yet, is that the default `jupyter-insert` is not getting overridden by the `R` method when I call `jupyter-eval-line-or-region` from an associated buffer. However the `R` method does correctly override the default version when executing a command directly in the REPL.

jackkamm:
On a somewhat different note -- I'd like to get Cask set up so that I can run tests, as Travis is failing since my latest commit (and also to avoid causing byte compile issues like 598359459c5cf803a92e89b3c9a11f16665a9876). However when I try to run `make test`, `make compile`, or `cask build`, I get errors like this:
```
jack@jaheira:~/src/emacs-jupyter (git)-[insert-html-browse-url]- % make compile
/usr/bin/cask emacs -Q -batch --eval "(progn (fset 'read-string (lambda (&rest _) \"y\")) (require 'zmq))"
Checking for compatible module binary to download
Contacting host: api.github.com:443
End of file while parsing JSON
make: *** [Makefile:34: zmq] Error 255
```
Any idea what might be going on here? I can't tell whether it's the fault of Cask, ZMQ, my emacs config, or something else, and my google fu is failing me here...

jupyter-mime.el
```diff
@@ -313,6 +313,18 @@ aligns on the current line."
         (setf (image-property image :ascent) 50)
         (force-window-update)))))
 
+(defun jupyter--insert-temp-file-browse-url (data)
+  (let ((file (make-temp-file "emacs-jupyter" nil ".html")))
+    (with-temp-file file (insert data))
+    (browse-url-of-file file)
+    ;; Give the external browser time to open the tmp file before deleting it
+    ;; based on mm-display-external
+    (let ((file file))
```
dzop: This `let` is also unnecessary, the lambda already captures the `file` variable of the first `let` form.

dzop:
> the `R` method is not overriding the default when calling `jupyter-eval-line-or-region` from an associated buffer

Can you place a call to

```emacs-lisp
(message "LANG %s" (jupyter-kernel-language jupyter-current-client))
```

just before https://github.com/dzop/emacs-jupyter/blob/a775ac29bd5ab5a903a4e9c03116281a9b299aed/jupyter-client.el#L889

and show what it returns. Also when evaluating results with `jupyter-eval-line-or-region`, the `text/plain` version is preferred so if the R kernel is also sending `text/plain` along with HTML that is probably why you are not seeing the `R` method being called.

jupyter-mime.el
```diff
@@ -466,11 +478,11 @@ width and height of the image."
 ;;; `jupyter-insert' method additions
 
 (cl-defmethod jupyter-insert ((_mime (eql :text/html)) data
-                              &context ((functionp 'libxml-parse-html-region)
```
dzop: Maybe we should instead prioritize the `jupyter-lang` context instead of removing this here so that whenever the `jupyter-lang` context is specified it means to override any other method and call that one first. The reason why you had to remove this context is because `(eql t)` takes precedence over `(jupyter-lang ...)`. If you change the 50 below

https://github.com/dzop/emacs-jupyter/blob/a775ac29bd5ab5a903a4e9c03116281a9b299aed/jupyter-client.el#L174

to 101 (100 is the priority of the `eql` specializer) and revert this change, the `R` method should be called first.

jackkamm: I couldn't get this to work -- it looks like priority is ignored in `&context` specializers. Consider the following minimal example:
```
(cl-defgeneric foo (bar)
  (ignore))

(cl-defmethod foo (bar &context ((list t) (head t)))
  (message "hello"))

(cl-defmethod foo (bar &context (t (eql t)))
  (message "goodbye"))

(foo "")
```
This should print `"goodbye"` since `eql` has priority 100 while `head` has priority 80. However, it prints out `"hello"` when I test it -- `cl-generic.el` seems to sort the methods by the order they're defined and not by the priority.

jackkamm:
> Can you place a call to
> 
> ```emacs-lisp
> (message "LANG %s" (jupyter-kernel-language jupyter-current-client))
> ```
> 
> just before
> 
> https://github.com/dzop/emacs-jupyter/blob/a775ac29bd5ab5a903a4e9c03116281a9b299aed/jupyter-client.el#L889

Doing this has no effect, because `jupyter--display-eval-result` is not called here; the result is being passed back through the `display-data` message type, not the `result` message type.

Inserting the call just above

https://github.com/dzop/emacs-jupyter/blob/ff0a228939f8c8bbc0fe0628f14ca406ee215241/jupyter-client.el#L978

prints out `LANG R` as expected.

jupyter-mime.el
jackkamm: Thanks, I've fixed this now.

jackkamm:
> Also when evaluating results with `jupyter-eval-line-or-region`, the `text/plain` version is preferred so if the R kernel is also sending `text/plain` along with HTML that is probably why you are not seeing the `R` method being called.

After modifying `jupyter-eval-string` so that message type `display-data` calls `jupyter--display-eval-result`, calling `jupyter-eval-line-or-region` from an associated buffer does print output, however as you say the plaintext is preferred, and in particular I get the message:
```
HTML widgets cannot be represented in plain text (need html)
```
Which is due to this line in the IRkernel: https://github.com/IRkernel/repr/blob/593305c080fd8b66587351457c10f0b09406d147/R/repr_htmlwidget.r#L118

jupyter-mime.el
dzop: Yes that seems to be the case. I misunderstood the priority slot of a generalizer.

After spending some time trying to understand it better, the priority of the generalizers factor in when there are multiple generalizers that can be used for a particular specializer. The highest priority generalizer gets a chance at creating an approximation of an object for use with the specializer. In your example, the `head` specializer has a generalizer that takes `(list t)` and generalizes it into 
`(car (list t))` and thats what the `head` specializer looks at to determine if the method is valid.

But there can be multiple generalizers for a specializer, whichever one returns a non-nil value is used by the specializer and the generalizers are tried in the order of their priority.

dzop:
OK I found the issue with the language specific methods not being called in the `display-data` callback, see 005f2e30bfc9c0e9a27d28e7ea6f61c04e24a28b. There should be no reason to modify any of the evaluation functions.

jupyter-mime.el
```diff
@@ -313,6 +313,17 @@ aligns on the current line."
         (setf (image-property image :ascent) 50)
         (force-window-update)))))
 
+(defun jupyter--insert-temp-file-browse-url (data)
```
dzop: Since you use this function in `jupyter-R.el` and is general enough that it could be used in other situations, it should really be made into a "public" function by removing the double dash and adding a doc string.

jackkamm: I've renamed to `jupyter-browse-url-in-temp-file` and added a docstring. Not sure it belongs in `jupyter-mime.el` still, but couldn't find a better place to put it, so I've left it there.

jupyter-client.el
```diff
@@ -977,9 +977,13 @@ to the above explanation."
...
-          (jupyter-display-current-buffer-reuse-window
-           :display-data nil (unless (jupyter-pop-up-frame-p :display-data)
-                               #'display-buffer-below-selected))))
+          (if (> (line-number-at-pos (point-max))
+                 jupyter-eval-short-result-max-lines)
+              (jupyter-display-current-buffer-reuse-window
+               :display-data nil (unless (jupyter-pop-up-frame-p :display-data)
+                                   #'display-buffer-below-selected))
+            (funcall jupyter-eval-short-result-display-function
+                     (buffer-string)))))
```
jackkamm: > OK I found the issue with the language specific methods not being called in the display-data callback, see 005f2e3. There should be no reason to modify any of the evaluation functions.

Thanks, now `jupyter-eval-line-or-region` calls the R-specific method as I'd like, and in particular calls `browse-url` on the result when appropriate. However, it also pops up an empty `*jupyter-display*` buffer (it's empty because of `jupyter--delete-script-tags`). So I added this snippet to use `jupyter-eval-short-result-display-function` when the output is short.

jupyter-mime.el
jackkamm: So it seems necessary to drop `&context` from the main `jupyter-insert` method for `text/html` output, unfortunately, because of the way priority works with `&context`. So I'm closing this thread. Please reopen if I'm wrong.

jupyter-client.el
dzop: I don't think having the `jupyter-eval-short-result-display-function` here is the right approach since its intended purpose is to display text results which is why its only used in `jupyter--display-eval-result`. If `data` contains an image to be displayed, `jupyter-insert` will insert a single space into the buffer and use the `display` text property to render the image (see `insert-image`). In this case, `jupyter-eval-short-result-display-function` would be called since there is only one line in the buffer containing a single space when really the buffer in which the image was inserted should be shown.

Since nothing is actually inserted into the buffer when calling `jupyter-browser-url-in-temp-file`, what you can do is to show the `*jupyter-display*` buffer only when `(/= (point-min) (point-max))` since it doesn't make sense to display an empty buffer in the first place. Could you also put a comment to explain why this is being done as well.

jackkamm: Good point, it looks terrible when `jupyter-eval-string` returns an image. I've fixed this as per your suggestion.

PR #120, opened 2019-05-19
Echo evaluation in REPL (fixes #71)

### LINKED ISSUE #71: Echo evaluated line
When calling a function like `jupyter-eval-line-or-region`, is there a way to make the line appear in the REPL like ESS does?

jackkamm:
This PR provides ESS/comint-style alternatives to `jupyter-eval-*`, that paste input/output in the REPL buffer, instead of displaying output  through a message or pop-up buffer.

Here's a screenshot:
jupyter-send-to-repl

For now, I've made `jupyter-send-to-repl-*` as separate functions from `jupyter-eval-*`, because that was the simplest to implement. Here's some options about what to do next:

0) Keep things as is, just add some keybindings for the new functions, and mention them in the Readme.
1) Make this behavior an option of `jupyter-eval-*` instead of creating separate functions.  The option could be set by user customization, or some other way (prefix-key? window arrangement?)
2)  Instead of separate functions, make `jupyter-eval-*` send output to **both** a pop-up buffer and the REPL at the same time.
    - Note: I initially tried going this route, using `jupyter-repl-insert` and `jupyter-repl-insert-prompt` to print the Input and a subsequent prompt, but printing the Output back to its associated cell seems trickier -- I remember getting an error about not being able to find the appropriate cell when I tried to use `jupyter-repl-append-output` just above this line: https://github.com/dzop/emacs-jupyter/blob/111d105b51e53bd254bdb59cb9d3fda1352c948a/jupyter-client.el#L974 I'll reproduce the specific error message if we decide it's better to go this route.

### DIFF
--- a/jupyter-client.el
+++ b/jupyter-client.el
@@ -1062,6 +1062,45 @@ representation of the results in the current buffer."
       (bounds-of-thing-at-point 'defun)
     (jupyter-eval-region beg end)))
 
+;; Alternative ESS-style evaluation that prints inputs/outputs in the
+;; REPL buffer
+
+(defun jupyter-send-to-repl-string (str)
+  "Paste STR to the repl and evaluate it there.
+If the repl's point is at the end, move it forward; otherwise,
+restore its position after evaluation."
+  (cl-check-type jupyter-current-client jupyter-kernel-client
+                 "Not a valid client")
+  (jupyter-with-repl-buffer jupyter-current-client
+    (let ((point-at-max (eql (point) (point-max))))
+      (save-excursion
+        (end-of-buffer)
+        (jupyter-repl-insert str)
+        (jupyter-repl-ret))
+      (when point-at-max
+        (end-of-buffer)))))
+
+(defun jupyter-send-to-repl-defun ()
+  "Paste the current function in the repl and evaluate it there."
+  (interactive)
+  (cl-destructuring-bind (beg . end)
+      (bounds-of-thing-at-point 'defun)
+    (jupyter-send-to-repl-string (buffer-substring beg end))))
+
+(defun jupyter-send-to-repl-line-or-region ()
+  "Paste the current line or region to the repl and evaluate it there."
+  (interactive)
+  (let* ((region (when (use-region-p)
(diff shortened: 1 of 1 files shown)


dzop:
I agree that we should not have separate evalutation functions just to copy the evaluated string to the REPL.

> send output to both a pop-up buffer and the REPL at the same time

I think this would work if we show a pop-up buffer conditionally. If the REPL buffer is invisible (visible) do (don't) pop-up a buffer.

> I remember getting an error about not being able to find the appropriate cell 

This is because `jupyter-repl-append-output` looks for the input cell associated with a request and this association is only made when code is evaluated from the REPL buffer, specifically whenever `jupyter-repl-finalize-cell` is called.

There should be no reason to use `jupyter-repl-append-output` though. We should arrange for the handler methods of the `jupyter-repl-client` to be called. Those methods will update the state of the REPL and are currently suppressed whenever `jupyter-eval-string` is called since `jupyter-eval-string` let binds the `jupyter-inhibit-handlers` variable.

I think we are going to need to define a new method to get the behavior we want. For example, we can have `jupyter-eval-string` be a method with its current definition being the default method and then
do something like the following for a `jupyter-repl-client`

```emacs-lisp
(cl-defmethod jupyter-eval-string (str &context (jupyter-current-client jupyter-repl-client)
                                       &optional _cb)
  (jupyter-with-repl-buffer jupyter-current-client
    (goto-char (point-max))
    (if (not (get-buffer-window nil 'visible))
        (let ((req (cl-call-next-method)))
          (jupyter-repl-finalize-cell req)
          (jupyter-repl-insert-prompt 'in)))
    (jupyter-repl-replace-cell-code str)
    (jupyter-send-execute-request jupyter-current-client)))
```

Note, this won't work as expected when the REPL buffer isn't visible since we still want the handlers to be called as well as the callback functions. Also the `jupyter-send-execute-request` method of the `jupyter-repl-client` does some processing of the REPL prompt that would be useful even when the REPL buffer isn't visible.

We will also need to define a command, `jupyter-eval-string-command` that just calls `jupyter-eval-string` so that it can still be bound to a key.

PR #121, opened 2019-05-22
Echo evaluation in the REPL

dzop:
This is continuing the discussion in #120 and should be a replacement for that pull request. Thanks to @jackkamm for his initial work on the issue.

This pull request does the following:

1. Converts `jupyter-eval-string` into a method.
2. In `jupyter-repl-client.el`, overrides the `jupyter-eval-string` method to obey the new variable `jupyter-repl-echo-eval-p` whenever the `jupyter-current-client` is a `jupyter-repl-client`.

The `jupyter-repl-echo-eval-p` variable controls whether or not evaluated input should be copied over to a REPL input cell. If this variable is `nil`, no copying happens and evaluation results are displayed as they are now, through pop-up buffers. When non-nil, input is copied to the REPL and evaluation results are also put into the REPL. However, in addition to appending results to the REPL, if the REPL buffer is not currently visible, the results are displayed using pop-up buffers.


jackkamm:
Thanks for taking this on and doing it right. I tested it and it works great. I'll go ahead and close #120 .

I only noticed 1 minor issue, specific to jupyter-R. Recall that jupyter-R opens some HTML outputs in an external browser. Sometimes this PR will cause the browser to be opened twice, when `jupyter-insert` is called twice (once in the REPL, once in the pop-up display).

Replacing `(jupyter-browse-url-in-temp-file data)` in `jupyter-R.el` with something like
```
      (unless (and jupyter-repl-echo-eval-p
                   (not (eql (current-buffer)
                             (slot-value jupyter-current-client 'buffer))))
        (jupyter-browse-url-in-temp-file data))
```
seems to fix the issue. That is, when `jupyter-repl-echo-eval-p`, only launch the browser when we are inserting in the REPL buffer -- this avoids opening it twice.

dzop:
Would it make sense to do something like the following in `jupyter-browse-url-in-temp-file`? This would prevent displaying the same file multiple times within the same minute. But it may be a problem if the user closes the browser within a minute and wants to run code that produces an identical html page. It doesn't seem that it would be that common of a situation though.

```emacs-lisp
(let* ((hash (sha1 (concat data (format-time-string "%H%M"))))
         (file (expand-file-name
                (concat "emacs-jupyter-" hash ".html")
                temporary-file-directory)))
    (unless (file-exists-p file)
      (with-temp-file file (insert data))
      (browse-url-of-file file)
      ;; Give the external browser time to open the tmp file before deleting it
      ;; based on mm-display-external
      (run-at-time
       60 nil
       (lambda ()
         (ignore-errors (delete-file file))))))
```

jackkamm:
I think this would be fine if we reduced the deletion time from 60 seconds to 5 seconds. I think it will occasionally happen that the user will close the browser, then decide she wants to see the same result again within a minute. But doing this within 5 seconds is a lot less likely, and 5 seconds should be more than enough time for the browser to start and open the file. (Testing this a few times, 1 second is enough for Firefox on my machine, at .5 seconds Firefox sometimes works, sometimes doesn't, and at .1 seconds Firefox will always give "File not found").

PR #135, opened 2019-06-22
133: added warning to README.org about possible conflicts with ob-ipython

cheremin:


README.org
```diff
@@ -398,6 +398,11 @@ example you can put the following in your configuration:
 (setq ob-async-no-async-languages-alist '("jupyter-python" "jupyter-julia"))
 #+END_SRC
 
+*** Issues with =ob-ipython=
```
dzop: Thanks for this.

Can you run `M-x org-fill-paragraph` (bound to `M-q`) on the two paragraphs so that they are formatted like the rest of the document.

README.org
```diff
@@ -398,6 +398,11 @@ example you can put the following in your configuration:
...
+If you already have =ob-ipython= installed, you /may/ experience issues with it conflicting with =emacs-jupyter= (e.g. [[https://github.com/dzop/emacs-jupyter/issues/133#issuecomment-502444999][this issue]). This is still not clearly investigated, but worth to keep in mind.
```
dzop: You are missing the ending bracket of the link.

```suggestion
If you already have =ob-ipython= installed, you /may/ experience issues with it conflicting with =emacs-jupyter= (e.g. [[https://github.com/dzop/emacs-jupyter/issues/133#issuecomment-502444999][this issue]]). This is still not clearly investigated, but worth to keep in mind. 
```

dzop: Also can we remove the part `This is still not clearly investigated, but worth to keep in mind.` and just give a brief explanation of the linked issue.

I took a look at `ob-ipython` and it has to do with its [Jupyter kernel support](https://github.com/gregsexton/ob-ipython#experimental-jupyter-kernel-support). Both `emacs-jupyter` and `ob-ipython` try to take ownership of `jupyter-LANG` source block names. So if both `ob-ipython` and `emacs-jupyter` are installed, `ob-ipython` will partially override what `emacs-jupyter` does to handle `jupyter-LANG` source blocks. The only way to suppress this is to remove the `org-mode-hook` [function](https://github.com/gregsexton/ob-ipython/blob/7147455230841744fb5b95dcbe03320313a77124/ob-ipython.el#L511) added by `ob-ipython` before an `org-mode` file is opened. Due to this, `ob-ipython` and `emacs-jupyter` are incompatible.

So you can actually replace this whole paragraph saying that `ob-ipython` must be uninstalled in order for `emacs-jupyter` to work properly since both packages define `jupyter-LANG` source blocks and, if you want, mention a little about why from the explanation above.

README.org
cheremin: Sure, np, here is it

README.org
cheremin: Sure, I have rephrase this block according to your investigation.

BTW: Thank you for investigation -- as for me, it is much simpler to just know tools are inherently incompatible, then guess that else could be the cause of incompatibility

dzop:
Thanks!

PR #147, opened 2019-07-13
Add jupyter-org-adjust-image-size.

Yevgnen:
If the `org` block outputs an image, and `org-image-actual-width` is a `list`, this PR provides customization to not to stretch the image if its width is smaller than `(car org-image-actual-width)`.

### DIFF
--- a/jupyter-org-client.el
+++ b/jupyter-org-client.el
@@ -73,6 +73,19 @@ automatically be shown if this is non-nil."
   :group 'ob-jupyter
   :type '(repeat string))
 
+(defcustom jupyter-org-adjust-image-size
+  t
+  "Try to best fit image output in the result block.
+
+If non-nil, and `org-image-actual-width' is set to a list, the
+image will not be stretched if its width is small than (car
+`org-image-actual-width'). This is done by inserting #+ATTR_ORG
+keyword above the file path.
+
+See the docstring of `org-image-actual-width' for more details."
+  :group 'ob-jupyter
+  :type 'boolean)
+
 (defconst jupyter-org-mime-types '(:text/org
                                    ;; Prioritize images over html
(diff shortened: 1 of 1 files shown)


jupyter-org-client.el
```diff
@@ -73,6 +73,19 @@ automatically be shown if this is non-nil."
   :group 'ob-jupyter
   :type '(repeat string))
 
+(defcustom jupyter-org-adjust-image-size
+  t
+  "Try to best fit image output in the result block.
+
+If non-nil, and `org-image-actual-width' is set to a list, the
+image will not be stretched if its width is small than (car
```
dzop:
```suggestion
image will not be stretched if its width is smaller than (car
```

jupyter-org-client.el
```diff
@@ -73,6 +73,19 @@ automatically be shown if this is non-nil."
...
+`org-image-actual-width'). This is done by inserting #+ATTR_ORG
```
dzop:
```suggestion
`org-image-actual-width'). This is done by inserting an #+ATTR_ORG
```

jupyter-org-client.el
```diff
@@ -868,6 +881,12 @@ the width or height of the image."
             (base64-decode-region (point-min) (point-max))))))
     (cl-destructuring-bind (&key width height &allow-other-keys)
         metadata
+      (if (and (null width)
+               jupyter-org-adjust-image-size
```
dzop: Can you mention this behavior in the docstring. You can just reference the documentation of `jupyter-org-adjust-image-size` somewhere in the last paragraph.

dzop:
Thanks! Looks good.

jupyter-org-client.el
Yevgnen: My poor English is killing me...

PR #188, opened 2019-10-04
Preserve content in current cell when set to echo eval

kevinjfoley:
Seems pretty straight forward but let me know if there's a better way of solving this.


jupyter-repl.el
```diff
@@ -1505,7 +1507,9 @@ value."
                 '(not :input-request))))
         (setq req (jupyter-send-execute-request jupyter-current-client
                     :code str
-                    :store-history jupyter-repl-echo-eval-p))))
+                    :store-history jupyter-repl-echo-eval-p))
+        (if cell-previous-code
```
dzop: I think just remove the check here. `jupyter-repl-cell-code` always returns a string, so its always going to be true anyways.

Sorry for taking so long to respond.

kevinjfoley: > Sorry for taking so long to respond.
No problem!

> I think just remove the check here. `jupyter-repl-cell-code` always returns a string, so its always going to be true anyways.

This is here for the case when `jupyter-repl-echo-eval-p` is `nil` and therefore `cell-previous-code` is `nil`.  Maybe it would be clearer if instead it checked `jupyter-repl-echo-eval-p`?

dzop: Oh yes, I see. It would be clearer using `jupyter-repl-echo-eval-p` so +1 from me.

We would then have 5 uses of `jupyter-repl-echo-eval-p` in the function, so I'm thinking we just refactor and do

```emacs-lisp
(if jupyter-repl-echo-eval-p
   ...
  ...)
```

But that can be left for another pull request if you are up for it.

kevinjfoley: 👍 

I'll make that change for now and open an issue to refactor

dzop:
Thanks!

PR #192, opened 2019-10-12
ob-jupyter: fixes #185 relative path to remote :session json

### LINKED ISSUE #185: Relative path to kernel json on TRAMP
If I start a kernel in the same directory as my org file like so:
```
jupyter console -f ./test.json --kernel julia-1.2
```

Then connect to the kernel through a relative path in the `:session` header like so:
```
#+begin_src jupyter-julia :session ./test.json
  x = 1+1
#+end_src
```

Then it works if the org file is on my local machine, but not if the org file is accessed remotely through TRAMP.

jackkamm:
Fixes https://github.com/dzop/emacs-jupyter/issues/185

### DIFF
--- a/ob-jupyter.el
+++ b/ob-jupyter.el
@@ -302,7 +302,8 @@ the host."
       (setq client
             (cond
              ((string-suffix-p ".json" session)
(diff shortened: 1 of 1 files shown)


dzop:
Thanks!

PR #193, opened 2019-10-13
Move whitespace to overlay prefix

arthurcgusmao:
Move the whitespace that was originally concatenated to the prefix variable `jupyter-eval-overlay-prefix` to the variable itself, so that it can be easily customized. For instance, in the case where one does not want any prefix, with the changes, it can be easily removed by setting the prefix variable to the empty string. Originally one would have to also modify the function `jupyter-eval-ov--propertize` to achieve the same result.


dzop:
Thanks!

PR #195, opened 2019-10-17
fix: show pop-up buffer when REPL not visible

arthurcgusmao:
The described behavior for when `jupyter-repl-echo-eval-p` is `t` and the REPL buffer is not visible was not working, i.e., no pop-up buffer was being show.


dzop:
Thanks for catching this.

arthurcgusmao:
You're welcome. Does the fact that the build was not successful have something to do with my changes? I have tested my commit with no issues on my machine.

dzop:
No, it looks like the failure was due to some networking issue on Appveyor not related to this change.

Issue #53, opened 2019-03-07
Support javascript output


GregorySchwartz:
I'm trying to use `hvega` which outputs `vega-lite` plots. The output is a javascript block. However, when running the source block I get no output and a warning: `Warning (emacs): No valid mimetype found (:application/javascript)`. This is when trying the barcode example in https://raw.githubusercontent.com/DougBurke/hvega/master/notebooks/VegaLiteGallery.ipynb

```
MESSAGE: (:iopub :display-data (:data (:application/javascript requirejs({paths:{vg:’https://cdn.jsdelivr.net/npm/vega@3.2.1/build/vega.min’,vl:’https://cdn.jsdelivr.net/npm/vega-lite@2.3.0/build/vega-lite.min’,vge:’https://cdn.jsdelivr.net/npm/vega-embed@3.5.2/build/vega-embed.min’},shim:{vge:{deps:[’vg.global’,’vl.global’]},vl:{deps:[’vg’]}}});define(’vg.global’,[’vg’],function(g){window.vega = g;});define(’vl.global’,[’vl’],function(g){window.vl = g;});var ndiv = document.createElement(’div’);ndiv.innerHTML = ’Awesome Vega-Lite visualization to appear here’;element[0].appendChild(ndiv);require([’vge’],function(vegaEmbed){vegaEmbed(ndiv,{"mark":"bar","data":{"values":[{"a":"A","b":28},{"a":"B","b":55},{"a":"C","b":43},{"a":"D","b":91},{"a":"E","b":81},{"a":"F","b":53},{"a":"G","b":19},{"a":"H","b":87},{"a":"I","b":52}]},"$schema":"https://vega.github.io/schema/vega-lite/v2.json","encoding":{"x":{"field":"a","type":"ordinal"},"y":{"field":"b","type":"quantitative"}},"description":"A simple bar chart with embedded data."}).then(function (result) { console.log(result); }).catch(function (error) { ndiv.innerHTML = ’There was an error: ’ + error; });});) :metadata nil))
```

dzop:
I don't know how practical it would be to print out the javascript since it looks like `vega-lite` produces minified `javascript` and Emacs will slow down to be nearly unusable in the presence of long lines (http://lists.gnu.org/archive/html/emacs-devel/2018-10/msg00471.html). This will be an issue for complex plots.

[...]

PR #198, opened 2019-11-06
Support javascript/html output

lane-s:
Addresses #53 by introducing a new node.js module called `web-renderer`. This simple module takes html input from stdin, renders it after waiting some time for javascript to run, takes a screenshot of the output, and then crops the screenshot to eliminate extra whitespace. The resulting PNG image is written to stdout.

`jupyter-org-client` has also been modified to support a new result type: web results. The idea is that passing `raw` for html content will render it using the new module. 

Many improvements need to be made before this is good, but I think the basic idea works.

A few notes:
 1. I changed the directory structure so that the `js` directory now has a Makefile that calls the `widget` Makefile and the `web-renderer` Makefile.
2. I made sure that I could build my branch using `cask package`, install using `package-install-file`, and then use the new feature. I've added the entire `web-renderer` module directory to the `Cask` file, so when Cask is used to build the package, `node_modules` is copied into the archive. This works for local installation, but it could be a problem if the archive built by Cask is going to go on Melpa. I think I might need to exclude `node_modules` and instead make sure `npm install` is run on the `web-renderer` module during the package installation or the first time the user tries to use the new feature.
3. The whole process of executing a code block and rendering a resulting image is quite slow. I suppose one way to speed it up would be to keep the node process running on a timeout, so multiple calls wouldn't need to start a new node process and a new headless browser every time. Once the user stops requesting images for a while, the process would stop on its own to avoid wasting system resources.
3. The image is trimmed a bit too close. It would be nice to trim it less or add a bit of a margin back after the trim.
4. We should look from Chrome in more default directories for various operating systems and also provide an option for the user to specify where their installation is. This could be a configuration variable in emacs that gets passed to the node module as an argument.
5. We could potentially also support taking the screenshot through Firefox using selenium.
6. This should also handle the `application/javascript` MIME type. I think we would just need to insert the code inside of a `<script>` tag inside the template we already have.

In the future, web results could also insert an xwidget in order to actually run javascript code in emacs as discussed in #135


lane-s:
I checked to see if connecting to a running chrome instance speeds things up at all, but it doesn't really seem to. I'm guessing that fetching the scripts isn't helping things along, so maybe there could be a mechanism to cache required scripts locally. It seems like even with simple html and no js, chrome takes a while to take the screenshot, so maybe it's not really worth trying to optimize.

PR #201, opened 2019-11-19
DO NOT MERGE: Update Julia REPL latex completion for upcoming changes

non-Jedi:
Recently made a PR to julia-mode to use `completion-at-point` for latex substitution: https://github.com/JuliaEditorSupport/julia-emacs/pull/82. This PR is the changes necessary for compatibility with that new implementation.

I didn't quite grok how `jupyter-completion-at-point` worked, so there may be details I'm missing, but from what I can tell, it's just duplicating the behavior of `julia-latexsub-completion-at-point`. Let me know if I'm missing something.

I'm just making this PR now so that I don't forget to do so later since I'm using this branch locally.

I did not update the README which uses `jupyter-indent-line` as defined for julia as an example because I didn't know what would be a good example to replace it with. There's also the question of whether `jupyter-indent-line` is worth maintaining since it was only used for julia.

### DIFF
--- a/jupyter-julia.el
+++ b/jupyter-julia.el
@@ -30,11 +30,7 @@
 (eval-when-compile (require 'subr-x))
 (require 'jupyter-repl)
 
-(declare-function julia-latexsub-or-indent "ext:julia-mode" (arg))
-
-(cl-defmethod jupyter-indent-line (&context (major-mode julia-mode))
-  "Call `julia-latexsub-or-indent'."
-  (call-interactively #'julia-latexsub-or-indent))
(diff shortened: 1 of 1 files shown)


non-Jedi:
I see now that there's logic to complete argument lists and symbols used as dictionary keys, so this PR is incorrect in what it removes. Will update when I get a chance.

MasonProtter:
Is this PR finished? If so, you may want to change the title.

non-Jedi:
@MasonProtter No. It's just a preemptive PR for when/if julia-mode is updated with https://github.com/JuliaEditorSupport/julia-emacs/pull/82, and it isn't working correctly anyways.

PR #208, opened 2019-12-19
fix: invalid message type

mackong:
for [almond](https://almond.sh/), when execute code like:
```
var m = Map("A" -> 1)
m += ("B" -> 2)
```
a *Invalid mesage type* error received, and almond kernel will exit.

### DIFF
--- a/jupyter-messages.el
+++ b/jupyter-messages.el
@@ -211,8 +211,9 @@ decoded string."
       (prog1 val
         (when-let* ((msg-type (and (listp val)
                                    (plist-get val :msg_type))))
-          (plist-put
-           val :msg_type (jupyter-message-type-as-keyword msg-type)))))))
+         (unless (string-empty-p msg-type))
+           (plist-put
+            val :msg_type (jupyter-message-type-as-keyword msg-type)))))))
 
 (defun jupyter-decode-time (str)
   "Decode an ISO 8601 time STR into a time object.


jupyter-messages.el
```diff
@@ -211,8 +211,9 @@ decoded string."
       (prog1 val
         (when-let* ((msg-type (and (listp val)
                                    (plist-get val :msg_type))))
-          (plist-put
-           val :msg_type (jupyter-message-type-as-keyword msg-type)))))))
+         (when (not (string-empty-p msg-type))
```
dzop: `(when (not` -> `(unless`

mackong: updated

dzop:
Actually I just realized that we kind of need a message to have a valid `:msg_type`. Not having a valid message type will most likely mess up some other parts of the code. The Jupyter messaging spec requires a valid message type for all messages, https://jupyter-client.readthedocs.io/en/stable/messaging.html.

Do you know why the `:msg_type` is an empty string? It should be one of the valid Jupyter message types, see `jupyter-message-types`.

https://github.com/dzop/emacs-jupyter/blob/de7af256a97813d6c0bd04e53aa36401cf3f8c33/jupyter-base.el#L71

Can you `(setq jupyter--debug t)`, evaluate the code that causes the error, and show here the messages that were printed in the `*Messages*` buffer.

nnicandro:
Closing this pull request since a message should have a valid message type. This is most likely an issue with the Almond kernel which has probably been fixed by now.

PR #211, opened 2020-01-08
jupyter-repl-traceback face: use a light background on light themes

astoff:
The previous value of this face used a dark background even on light themes, making
error messages hard to read.

This is how the proposed new setting looks like on Emacs's default theme:

image


dzop:
Thanks.

PR #230, opened 2020-03-14
Fix #219 REPL freezing in Emacs-27

### LINKED ISSUE #219: REPL crashes on Emacs 27
After switching to Emacs 27.0.60, I started to experience crashes/super slowness when interacting with a Jupyter REPL buffer. After pressing `C-g` several times to quit the crashing state, the following was output in the messages buffer:
```
Error during redisplay: (jit-lock-function 180) signaled (quit)
Error during redisplay: (jit-lock-function 181) signaled (quit)
Error during redisplay: (jit-lock-function 182) signaled (quit)
Error during redisplay: (jit-lock-function 186) signaled (quit)
Error during redisplay: (jit-lock-function 187) signaled (quit)
Error during redisplay: (jit-lock-function 193) signaled (quit)
Error during redisplay: (jit-lock-function 194) signaled (quit)
Error during redisplay: (jit-lock-function 180) signaled (beginning-of-buffer)
```

So it must have something to do with displaying the buffer, and, more precisely, with `font-lock`. Indeed, when I disable `font-lock` (such as when using the hotfix snippet below), the slowness/crashing stops. It doesn't seem to be anything particular to my Emacs configurations since it was working normally before. How should we fix it?

```lisp
;; Hotfix
(add-hook 'jupyter-repl-mode-hook
          (lambda () (font-lock-mode 0)))
```

EDIT
Maybe something to do with [this new feature](https://github.com/emacs-mirror/emacs/blob/master/etc/NEWS.27#L718-L726)?
I tried setting the variable `jit-lock-antiblink-grace` to `nil` but the behavior remained the same, with the same errors.

arthurcgusmao:
Fixed the issue of Emacs 27 freezing when sending code to a REPL buffer
with font-lock mode activated. The issue started to happen due to a
refactor of part of the code in `font-lock-default-fontify-region` to `font-lock-fontify-region`.

See https://github.com/dzop/emacs-jupyter/issues/219

### DIFF
--- a/jupyter-repl.el
+++ b/jupyter-repl.el
@@ -1863,7 +1863,7 @@ VERBOSE has the same meaning as in
   (let (fld frf sff spf comment)
     (jupyter-with-repl-lang-buffer
       (setq fld font-lock-defaults
-            frf (or font-lock-fontify-region-function #'ignore)
+            frf (or #'font-lock-fontify-region #'ignore)
             sff (or font-lock-syntactic-face-function #'ignore)
             spf (or syntax-propertize-function #'ignore)
             comment comment-start))


arthurcgusmao:
Fix not working and not implemented as it ideally should, better to go for a different approach.

PR #236, opened 2020-04-02
Work with proposed updates to julia-mode LaTeX completion

non-Jedi:
https://github.com/JuliaEditorSupport/julia-emacs/pull/100 changes
LaTeX completion in julia-mode to work by a combination of
completion-at-point-function and abbrev-mode. This is the necessary
change to give full-featured LaTeX completion with emacs-jupyter julia
buffers with the PR in its current form.

Unlike #201, this is actually working. Obviously shouldn't be merged until/if https://github.com/JuliaEditorSupport/julia-emacs/pull/100 is merged, but since I've already done the work, I thought it worthwhile to publish here.


non-Jedi:
The LaTeX symbol completion rework has been merged into julia-mode, so this needs to be fixed now. Unfortunately, this PR isn't working as is. It seems like something is clearing out `completion-at-point-functions` because neither of `julia-mode-latexsub-completion-at-point-around` nor `julia-mode-latexsub-completion-at-point-before` are even getting called.

Quickly skimming through the code in `jupyter-repl.el`, I'm not seeing why this would be the case though. @nnicandro, any guidance on why adding things to `completion-at-point-functions` to run before `jupyter-completion-at-point` wouldn't work?

nnicandro:
I'm not sure what could be going on. The completion code hasn't changed in a long time. Have you tried using something like `add-variable-watcher` on `completion-at-point-functions` to check what function is changing the value of that variable, e.g. you can put a call to `backtrace` in the `watch-function`. Note there is also `remove-variable-watcher` to undo the watcher.

PR #238, opened 2020-04-03
Add jupyter-describe function for interactive inspect

### LINKED ISSUE #199: Interactive jupyter-inspect with prompt
Would be great if there was a function hooking into the `jupyter-inspect` machinery that would prompt for a symbol and display the results for that symbol.

My vision for this is that you could also hook into the completion machinery and have an interface similar to `describe-function`/`describe-variable` but for Julia symbols.

I'll submit a PR if interest.

non-Jedi:
jupyter-describe implements an interface (with completion) similar to
emacs builtin describe-function for interacting with jupyter kernels.

Closes #199

This was actually easier than I expected, so kudos for the sane architecture. :)

### DIFF
--- a/jupyter-client.el
+++ b/jupyter-client.el
@@ -946,24 +946,32 @@ client local variable.
 
 Methods that extend this generic function should
 `cl-call-next-method' as a last step."
-  (cl-check-type jupyter-current-client jupyter-kernel-client
-                 "Need a client to read an expression")
   (let* ((client jupyter-current-client)
          (jupyter--read-expression-history
           (jupyter-get client 'jupyter-eval-expression-history)))
-    (minibuffer-with-setup-hook
-        (lambda ()
-          (setq jupyter-current-client client)
-          (add-hook 'completion-at-point-functions
-                    'jupyter-completion-at-point nil t)
-          (add-hook 'minibuffer-exit-hook
-                    'jupyter--teardown-minibuffer nil t))
-      (prog1 (read-from-minibuffer
-              (format "Eval (%s): " (jupyter-kernel-language client))
-              nil read-expression-map
-              nil 'jupyter--read-expression-history)
-        (jupyter-set client 'jupyter-eval-expression-history
-                     jupyter--read-expression-history)))))
+    (prog1
+        (jupyter--read-with-completion
+         client "Eval (%s): " jupyter--read-expression-history)
+      (jupyter-set client 'jupyter-eval-expression-history
+                   jupyter--read-expression-history))))
+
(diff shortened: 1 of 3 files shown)


jupyter-client.el
```diff
@@ -946,24 +946,32 @@ client local variable.
...
+(defun jupyter--read-with-completion (client prompt &optional history)
+  "Read an expression using CLIENT for completion.
+The expression is read from the minibuffer with PROMPT and expression
+history HISTORY."
+  (cl-check-type client jupyter-kernel-client
+                 "Need a client to read an expression")
+  (minibuffer-with-setup-hook
+      (lambda ()
+        (setq jupyter-current-client client)
+        (add-hook 'completion-at-point-functions
```
non-Jedi, line 967: This should probably use `completing-read` instead. Any idea why it was written this way instead?

dzop: I copied it over from `read--expression` called by `eval-expression` (`M-:`). My intention with `jupyter-read-expression` is to parallel `read--expression`. Also `completing-read` is more for completing collections right? `jupyter-read-expression` is supposed to read a line of code for evaluation from the minibuffer while also allowing completion from the kernel within the line.

Although the documentation for `jupyter-read-expression` does say `for completion` instead of something more approriate like `for evalution` so its purpose isn't that clear.

jupyter-client.el
```diff
@@ -946,24 +946,32 @@ client local variable.
...
+    (prog1
+        (jupyter--read-with-completion
+         client "Eval (%s): " jupyter--read-expression-history)
+      (jupyter-set client 'jupyter-eval-expression-history
+                   jupyter--read-expression-history))))
+
+(defun jupyter--read-with-completion (client prompt &optional history)
+  "Read an expression using CLIENT for completion.
+The expression is read from the minibuffer with PROMPT and expression
+history HISTORY."
```
dzop, line 961:
```suggestion
HISTORY."
```

jupyter-client.el
non-Jedi: Hm. That makes sense. So for this new purpose I'm putting this code to, using `completing-read` probably makes more sense, but it doesn't make sense in the context of `jupyter-read-expression` which is likely to complete multiple symbols. I think this is an acceptable hack though until I finish a `completing-read` based implementation?

non-Jedi: In either case, I think it probably makes more sense to just refactor `jupyter-read-expression` to take optional parameters rather than define an internal `jupyter--read-with-completion`. :/

PR #248, opened 2020-04-23
Require eieio for jupyter-channel defclass

dakra:
Recently I switched to the Emacs `native-comp` branch, and while the missing require apparently didn't really hurt before, after native compiling I got an error that it couldn't find variable jupyter-channel when loading jupyter.
This seems to fix it.

### DIFF
--- a/jupyter-channel.el
+++ b/jupyter-channel.el
@@ -26,6 +26,8 @@
 
 ;;; Code:
 
+(require 'eieio)
+
 (defclass jupyter-channel ()
   ((type
     :type keyword


dakra:
Friendly ping as this bug also came up on the Emacs bug tracker recently https://debbugs.gnu.org/cgi/bugreport.cgi?bug=43969

ricardog:
I filed https://debbugs.gnu.org/cgi/bugreport.cgi?bug=43969 and can confirm this patch fixed the problem locally.

Thanks.

nnicandro:
Sorry for being so late on this, thanks.

PR #262, opened 2020-05-19
parse `DataFrame` from evaluating a Julia cell as an Org table

FelipeLema:
[Julia DataFrames](https://juliadata.github.io/DataFrames.jl/stable/) is a [famous package](https://juliaobserver.com/packages/DataFrames) in the Julia universe (think Python's pandas).

This will parse the html representation of a `DataFrame` after evaluating a cell. I tried to keep things maintainable, although recursing through html is always cumbersome.

I don't know where/if I should add a unit test for this. Is there a guide for this?

Also, I think I might've missed the way headers should be passed to `jupyter-org-scalar`. If passing headers is not part of this function, then I guess we're ok.

### DIFF
--- a/jupyter-julia.el
+++ b/jupyter-julia.el
@@ -248,35 +248,35 @@ TABLE-AS-HTML-STR is the jupyter html representation of a DataFrame, like
 Returns \(cons 'ok org-table-representing-dataframe\), nil otherwise."
   (when (not (null table-as-html-str))
     (when-let* ((e (with-temp-buffer
-                       (insert table-as-html-str)
-                       (libxml-parse-html-region (point-min) (point-max))))
-                  (headers-and-rows (pcase e
-                                      (`(html , _html-attribs
-                                                (body , _body-attribs
-                                                        (table ((class . "data-frame"))
-                                                               . ,rest)))
-                                       rest)))
-                  (first-tr (pcase (car headers-and-rows)
-                              (`(thead ,_attr . ,rest)
-                               (car rest))))
-                  (th-nodes (pcase first-tr
-                              (`(tr ,_attrib . ,ths)
-                               ths)))
-                  (col-names
-                   (mapcar
-                    'caddr
-                    (--drop-while
-                     (pcase it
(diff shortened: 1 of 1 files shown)


FelipeLema:
this is no longer valid as-is because of https://github.com/JuliaData/DataFrames.jl/pull/2429

FelipeLema:
I just tried this with a recent julia setup because HTML representation (which is the one being parsed) hasn't changed.

nnicandro:
How stable is the HTML produced by the DataFrame package? Has there been a lot of changes to it recently? I would prefer to make this addition knowing that the HTML produced is not in flux so that we don't have to keep up with changes to it.

Also `emacs-jupyter` doesn't use the `dash` library so can you change those `--drop-while` calls to something like

```emacs-lisp
(nthcdr
 (cl-position-if-not
  (lambda (x)
    (pcase x
      (`(th ,_attrib . ,col-name)
       (null col-name))))
  th-nodes)
 th-nodes)
```

Regarding a test. I think we should have one. RIght now all the tests only work with Python kernels since that is the default kernel on any Jupyter installation. I'll have to make it easier to write tests using kernels other than Python. Below is an example of a current test that can be adapted for this case. The `jupyter-org-test-src-block` macro assumes a Python kernel so it will have to be re-written so that one can optionally provide a different kernel to use. That's my suggested approach.

https://github.com/nnicandro/emacs-jupyter/blob/b87ecf0cd3697bcdc5180fc5935786f2d282538d/test/jupyter-test.el#L1976-L1986

`jupyter-org-scalar` doesn't do anything with source block headers so I don't think they are needed for this case. Would there be any header arguments that could be useful to consider here?

FelipeLema:
I know that DataFrames.jl's html printing is very stable, since I've been using them until rather recently.

I wrote this method before a major re-write they did some months ago. It seems that it was just behind-the-scenes code since this method kept on working Ok.

Thanks for the thorough answer. I'll address the other points in the next days

FelipeLema:
I've addressed the comments and this is ready to be squash-pulled

PR #293, opened 2020-10-16
Handle ANSI escape characters in results of src-block to fix #197

### LINKED ISSUE #197: ANSI escape characters not rendering correctly
See #56 for more context.

As @dzop requested, here's some additional debugging output.

In an org-mode buffer I ran `eval-expression` and ran:
```elisp
(with-temp-buffer
  (org-mode)
  (insert ": �[31mhello�[39m")
  (font-lock-ensure)
  (text-properties-at 8))
```

which returned

```elisp
(face org-code wrap-prefix "" line-prefix "")
```

I created an elisp file (with the proper escape character which isn't displayed here):

```elisp
(push "~/.emacs.d/elpa/26.3/develop/jupyter-20190917.1642" load-path)
(push "~/.emacs.d/elpa/26.3/develop/simple-httpd-20190110.1505" load-path)
(push "~/.emacs.d/elpa/26.3/develop/websocket-20190621.54" load-path)
(push "~/.emacs.d/elpa/26.3/develop/org-plus-contrib-20190909" load-path)
(require 'org)
(require 'jupyter-org-client)
(with-temp-buffer
  (org-mode)
  (insert ": [31mhello[39m")
  (font-lock-ensure)
  (prin1 (text-properties-at 8)))
```

And ran it with `emacs -Q -batch -l /tmp/ansi-test.el`. emacs exited with status `0` but otherwise didn't produce any output.

haldai:
Adding a function in `ob-jupyter.el` to fix [this issue](https://github.com/nnicandro/emacs-jupyter/issues/197#issuecomment-710561406).

### DIFF
--- a/ob-jupyter.el
+++ b/ob-jupyter.el
@@ -730,11 +730,19 @@ mapped to their appropriate minted language in
         (goto-char pos)
         (ansi-color-apply-on-region (point) (org-babel-result-end))))))
 
+(defun org-babel-jupyter-handle-result-ansi-escapes ()
+  "Remove ANSI escapes from Jupyter src-block results in the current buffer."
+  (org-babel-map-src-blocks nil
+    (when (org-babel-jupyter-language-p lang)
+      (goto-char (org-babel-where-is-src-block-result))
+      (ansi-color-apply-on-region (point) (org-babel-result-end)))))
+
 ;;; Hook into `org'
 
 (org-babel-jupyter-aliases-from-kernelspecs)
 (add-hook 'org-export-before-processing-hook #'org-babel-jupyter-setup-export)
 (add-hook 'org-export-before-parsing-hook #'org-babel-jupyter-strip-ansi-escapes)
+(add-hook 'org-babel-after-execute-hook #'org-babel-jupyter-handle-result-ansi-escapes)
 
 (provide 'ob-jupyter)
 


nnicandro:
The ANSI escapes should already be handled by this code

https://github.com/nnicandro/emacs-jupyter/blob/b87ecf0cd3697bcdc5180fc5935786f2d282538d/jupyter-org-client.el#L583-L603

Your code is basically repeating what the above code does. There must be something else going on that causes the ANSI sequences to not be handled in the issue you reference. I'm not sure what that is yet, but when I run your example

```emacs-lisp
#+begin_src jupyter-julia :session jl1
colors = [:red, :blue, :green, :yellow, :white]
for c in colors
    printstyled("hello\n", color=c)
end
#+end_src
```

I get properly highlighted output.

dangom:
I have the same issue, and the proposed fix also doesn't work as is.
It seems the problem stems from the output text being printed after org-babel-execute and it's hooks have finished running.

In my case adapting the hook in the proposed PR to the following fixes the issue:

```elisp
(add-hook 'org-babel-after-execute-hook #'(lambda () (run-with-timer 0.2 nil #'org-babel-jupyter-handle-result-ansi-escapes)))
```

But as you see, this is not a super clean solution.

Issue #287, opened 2020-09-04
Emacs hangs when unicode entereed to Jupyter REPL (Julia)


dangirsh:
Julia allows unicode characters in variables names, like `u₀`. If I enter or paste these into a Jupyter REPL, my Emacs hangs at 100% CPU and I have to kill it externally with `pkill -9 emacs`. 

It's not fully reproducible, but happens often enough that I avoid the REPL altogether. There are no issues with evaluating code that uses unicode characters in Org src blocks, or in buffers associated with the kernel. This seems to be specific to Jupyter REPL. 

System info in the "details" below:

<details>

```
emacs   version    28.0.50
        features   XPM JPEG TIFF GIF PNG RSVG CAIRO SOUND DBUS GSETTINGS GLIB NOTIFY INOTIFY LIBSELINUX GNUTLS LIBXML2 FREETYPE HARFBUZZ ZLIB TOOLKIT_SCROLL_BARS GTK3 X11 XDBE XIM MODULES THREADS PDUMPER LCMS2
        build      Aug 05, 2020
        buildopts  (--with-modules)
        windowsys  x
        daemonp    server-running
doom    version    2.0.9
        build      grafted, HEAD -> develop, origin/develop, origin/HEAD 00a8f75 2020-08-05 03:32:21 -0400
        dir        ~/.doom.d/
system  type       gnu/linux
        config     x86_64-pc-linux-gnu
        shell      /bin/bash
        uname      Linux 5.4.0-42-generic #46-Ubuntu SMP Fri Jul 10 00:24:02 UTC 2020 x86_64
        path       (~/.cargo/bin ~/.nix-profile/bin ~/.guix-profile/bin /usr/local/sbin /usr/local/bin /usr/sbin /usr/bin /sbin /bin /usr/games /usr/local/games /snap/bin /usr/local/go/bin ~/emacs-src-new/lib-src/)
config  envfile    envvar-file
        elc-files  0
[...]

PR #308, opened 2021-01-17
Don't add the font-lock-multiline property to finalized cell text

nnicandro:
This is a fix for #219.

Looking further into the issue and with the help of the reproduction in https://github.com/nnicandro/emacs-jupyter/issues/219#issuecomment-641487092 it seems that property added to the cell text causes an infinite loop in the `font-lock-extend-region-functions` handling during fontification whenever the input is long.

### DIFF
--- a/jupyter-repl.el
+++ b/jupyter-repl.el
@@ -739,9 +739,7 @@ Place `point' at `point-max'."
     ;; cell or after any continuation prompts.  See
     ;; `jupyter-repl-insert-prompt'.
     (remove-text-properties beg (point) '(rear-nonsticky))
-    ;; font-lock-multiline to avoid improper syntactic elements from
-    ;; spilling over to the rest of the buffer.
-    (add-text-properties beg (point) '(read-only t font-lock-multiline t))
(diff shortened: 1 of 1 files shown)


kleinschmidt:
This does seem to fix up the issue where sending large chunks to the REPL causes it to hang (and probably the unicode issues from #287) but there's no font-locking with big chunks.  To be clear I can live with that very happily but just want to report in case it's not expected behavior.

nnicandro:
There should be font-locking so it's definately a bug.  Does the code that your testing have string characters in it? If you type in `function "` does the font-locking disappear on `function`?

kleinschmidt:
Yup, that it does.

kleinschmidt:
I'll keep an eye out for other examples this week...

nnicandro:
I just pushed a commit to this branch that should fix the problem of no font-locking when string characters are present if you want to give it a shot.

kleinschmidt:
Yeah that seems to do it.  I'll keep working with this and see if other issues come up...

PR #320, opened 2021-03-30
Update jupyter--encode for new Emacs JSON API.

ydl7:
The old Emacs JSON API (for versions <= 27.2 at time of commit) was
structured in a way that replacing the symbol-function of json-encode
worked to get custom encoding for certain object types used by
emacs-jupyter. In the later API, the function json--print plays the
recursive role of json-encode, and the old method does not work. The 
code here works in either version, and is simplified so that any future 
updates to the API are easier to accomodate.

### DIFF
--- a/jupyter-messages.el
+++ b/jupyter-messages.el
@@ -152,38 +152,34 @@ If PART is a list whose first element is the symbol,
 is non-nil.  If it is nil, then set the list's second element to
 the result of calling `jupyter--encode' on the third element and
 return the result."
-  (if (eq (car-safe part) 'message-part)
-      (cl-destructuring-bind (_ encoded-rep decoded-rep) part
-        (or encoded-rep (setf (nth 1 part) (jupyter--encode decoded-rep))))
-    (cl-letf (((symbol-function 'json-encode)
-               (lambda (object)
-                 (cond ((memq object (list t json-null json-false))
-                        (json-encode-keyword object))
-                       ((stringp object)      (json-encode-string object))
-                       ((keywordp object)
-                        ;; Handle `jupyter-message-type'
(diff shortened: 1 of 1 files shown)


jupyter-messages.el
```diff
@@ -152,38 +152,34 @@ If PART is a list whose first element is the symbol,
...
-                               (cl-every #'integerp object))
-                          (jupyter-encode-time object))
-                         (t (json-encode-list object))))
-                       (t                     (signal 'json-error (list object)))))))
-      (encode-coding-string
-       (cond
-        ((stringp part) part)
-        (t (json-encode part)))
-       'utf-8 t))))
+  (let ((json-recursive-encoder-sym (if (fboundp 'json--print) 'json--print 'json-encode)))
```
nnicandro, line 155: Can you move the `if` call outside of the function and into a `defvar`?  I think that would be better than checking for `json--print` every time a message gets encoded.

nnicandro:
Thanks for this.  Much better than what I had.

nnicandro:
Could you also rebase against the master branch.  I've updated the Travis config so the tests should be working again.

jupyter-messages.el
ydl7: I could, but I think the "right" choice in terms of behaviour is checking every time since in principle the user could load the new or old JSON APIs during an Emacs session. Although this is an (extreme) edge case, since the check is inexpensive (especially compared to the rest of the function; note this check is only done once when the JSON encoder is called, not recursively) I have a weak preference to keep it this way.

Let me know if you still want the check moved to a `defvar` in light of this, happy to do so then.

nnicandro: Your right, it only happens once per message encoding so it's not a big deal.

PR #364, opened 2021-11-27
Update `dev`, remove `init` target in Makefile (no `--dev` option)

xor-xor:
Since [this commit](https://github.com/cask/cask/commit/a8a2f72bef5e9a8ec62b6f1a0e52a8caf554570a), Cask doesn't provide `--dev` option anymore. Also, given that `dev` target looks like `init` now, I guess that the latter is not needed.


nnicandro:
Thanks for this!

PR #372, opened 2021-12-19
Add optional image width limiting for REPL

xzz53:


nnicandro:
Can you name the variable `jupyter-image-max-width` instead of `jupyter-repl-image-max-width`. The `jupyter-mime.el` file is not in the namespace of `jupyter-repl` so I would prefer there not to be any `jupyter-repl-*` variables defined in it.

xzz53:
> Can you name the variable `jupyter-image-max-width` instead of `jupyter-repl-image-max-width`. The `jupyter-mime.el` file is not in the namespace of `jupyter-repl` so I would prefer there not to be any `jupyter-repl-*` variables defined in it.

Done.

PR #377, opened 2022-01-11
Support parsing and highlighting REPL input cells with tree-sitter

ubolonton:
This will fix https://github.com/emacs-tree-sitter/elisp-tree-sitter/issues/78.

The design is described in https://github.com/emacs-tree-sitter/elisp-tree-sitter/pull/202.

For Python, it can be enabled by evaluating `(tree-sitter-enable 'python)` in the REPL buffer.
It should also work for other languages. I tested it with Scala (Almond).

### DIFF
--- a/jupyter-repl.el
+++ b/jupyter-repl.el
@@ -423,6 +423,11 @@ interpreted as `in'."
        ;; buffer modified anymore.  This is also an indicator for when undo's
        ;; can be made in the buffer.
        (set-buffer-modified-p nil)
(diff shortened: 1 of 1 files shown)


nnicandro:
Closing this as Emacs now has its own tree-sitter support built-in although I'm not sure if some support is needed in Emacs-Jupyter to use it or it can already be used. If someone finds that they cannot use the built-in Emacs support for tree-sitter when highlighting REPL cells please feel free to open an issue.

Sorry for not getting to this sooner.

PR #388, opened 2022-02-25
Discard stderr from jupyter shell command

indigoviolet:
In some situations, `jupyter kernelspec list --json` produces stderr warnings like

```
[ListKernelSpecs] WARNING | Native kernel (python3) is not available
```
which leads to a JSON parsing error

This change discards the stderr.


nnicandro:
Thanks for this but I prefer #418.  I think discarding stderr completely in `jupyter-command` is a bit overkill. There may be a case for looking at the stderr messages if they do happen in other calls to `jupyter-command`. Although there aren't any such cases at the moment.

PR #389, opened 2022-02-27
Sort available kernelspecs; guess with ^; use guessing in jupyter-run-repl

### LINKED ISSUE #371: `jupyter-run-repl` starts wrong kernel when called interactively
Let's say that we have two kernels: `numpy_tmp` and `tmp`. If we'll call `jupyter-run-repl` interactively with an intention to run `tmp`, we'll actually get `numpy_tmp` instead - and this happens despite the fact that `jupyter-completing-read-kernelspec` returns an exact name (`tmp`), so there's no ambiguity here (which, in contrast, needs to be resolved in case of non-interactive calls).

Basically, this issue seems to be an easy one to solve, as it boils down to the use of `jupyter-find-kernelspecs` function. It appears in `jupyter-run-repl` first, but is skipped when `jupyter-run-repl` is called interactively, which totally makes sense. Then it appears again, deeper down in `jupyter-start-new-kernel`, wrapped in `jupyter-guess-kernelspec`, but this time it is not guarded by `called-interactively-p` - and this is precisely where this bug comes from.

One may think: "ok, so let's just change this second call from `jupyter-guess-kernelspec` to `jupyter-get-kernelspec`, as we have an exact name, so there's no need for guessing" - but doing this will probably break things if someone is relying on the behavior described in `jupyter-start-new-kernel`'s docstring:

> KERNEL-NAME is the name of the kernel to start.  **It can also be the prefix of a valid kernel name, in which case the first kernel in `jupyter-available-kernelspecs' that has KERNEL-NAME as a prefix will be used**. 

Any thoughts on how to proceed here?

EDIT: on the other hand, `jupyter-run-repl` may be simply changed to require an exact name of an existing kernel and signal an error otherwise, but then again, it may break things if someone is relying on this behavior:

> Run a Jupyter REPL connected to a kernel with name, KERNEL-NAME. **KERNEL-NAME will be passed to `jupyter-find-kernelspecs' and the first kernel found will be used to start the new kernel**.

xor-xor:
Fixes #371.

Turns out that there's no need to wrap `jupyter-run-repl`'s `KERNEL-NAME` in `^$`, as only `^` is needed when kernelspecs are sorted by names - and I think that this is actually closer to the original semantics, i.e. being able to specify just `python` instead of `python3` or `julia` instead of `julia-X.Y`.

With this approach, there's also no need for this extra call to `jupyter-find-kernelspecs` in `jupyter-run-repl` (and by extension, for this "fragile" check if `jupyter-run-repl` has been called interactively).

So, probably the most "controversial" change in this PR is "starting with `NAME`" instead of "matching `NAME`" in `jupyter-guess-kernelspec` - but I think that [the original idea for this fix](https://github.com/nnicandro/emacs-jupyter/issues/371#issuecomment-1038514708) was actually more "invasive" ;)

### DIFF
--- a/jupyter-kernelspec.el
+++ b/jupyter-kernelspec.el
@@ -45,10 +45,10 @@
   "Get the available kernelspecs.
 Return an alist mapping kernel names to (DIRECTORY . PLIST) pairs
 where DIRECTORY is the resource directory of the kernel and PLIST
-is its kernelspec plist.  The alist is formed by parsing the
-output of the shell command
+is its kernelspec plist.  The alist is formed by parsing and
+sorting the output of the shell command
 
-    jupyter kernelspec list
+    jupyter kernelspec list --json
 
 By default the available kernelspecs are cached.  To force an
 update of the cached kernelspecs, give a non-nil value to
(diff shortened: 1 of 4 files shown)


nnicandro:
Thanks for this!

PR #408, opened 2022-08-16
jupyter-repl-mode handle yank-excluded-properties t

tgbugs:
The value of `yank-excluded-properties' can be t, in which case it is not
possible to `remq' 'field, causing an error. This fix will allow the mode to
start, but does not handle the fact that the 'field property is still stripped.


tgbugs:
A better fix might include the following.
```emacs-lisp
(setq-local yank-excluded-properties
              (remq 'field 
                    (or
                     (and (listp yank-excluded-properties)
                          yank-excluded-properties)
                     ;; if the user sets `yank-excluded-properties' to t we cannot recover
                     ;; the default list of properties, so it is reproduced here for consistency
                     '(category field follow-link fontified font-lock-face help-echo
                       intangible invisible keymap local-map mouse-face read-only yank-handler))))
```

PR #418, opened 2022-11-14
Fix available-kernelspecs failing from warning output

### LINKED ISSUE #301: Emacs 27 json-read: JSON readtable error: 76
Trying to move from ob-ipython to emacs-jupyter, I am having trouble getting it to run. 

My main environment that I wish to work from is Anaconda on a Linux machine. 
I do a minimal install of emacs-jupyter with use-package:

```
(use-package emacs-jupyter
  :ensure t
  )
```

Emacs claims it installed jupyter just fine, but when trying to run the REPL, I get the error message:

```
json-read: JSON readtable error: 76 
```

...rather cryptic, I think. I have not een able to find out whether emacs-jupyter finds the Anaconda jupyter kernels and interpreters, but org-babel also does not understand any jupyter-XXX named kernels even though I did

```
  (org-babel-do-load-languages
   'org-babel-load-languages
   '((R . t) 
     (python . t) 
     ;;(ipython . t) 
     (latex . t)
     (ditaa . t)
     (shell . t)
     (jupyter . t)
     ))

  (org-babel-jupyter-override-src-block 'python)
```
...with no luck. 
What am I doing wrong?

ahmed-shariff:
Would fix #301 

The PR limits the logs to `ERROR` so that the function (`jupyter-available-kernelspecs`) can still work if there is a warning. If `ERROR` or above is printed, probably it's a good idea for the function to fail?

### DIFF
--- a/jupyter-kernelspec.el
+++ b/jupyter-kernelspec.el
@@ -61,7 +61,7 @@ each DIRECTORY will be a remote file name."
     (or (and (not refresh) (gethash host jupyter--kernelspecs))
         (let ((specs (plist-get
                       (jupyter-read-plist-from-string
-                       (or (jupyter-command "kernelspec" "list" "--json")
+                       (or (jupyter-command "kernelspec" "list" "--json" "--log-level" "ERROR")
                            (error "Can't obtain kernelspecs from jupyter shell command")))
                       :kernelspecs)))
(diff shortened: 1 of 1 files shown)


nnicandro:
Thanks!

PR #421, opened 2022-11-28
fix for #395 

### LINKED ISSUE #395: error in server REPL after connection drops: `(wrong-number-of-arguments (3 . 3) 4)`
Occasionally, when I hit Ret in a server REPL after the connection to the server has dropped (port forwarded over SSH), I get the following error (debugger output):

```
Debugger entered--Lisp error: (wrong-number-of-arguments (3 . 3) 4)
  #f(compiled-function (client req msg) #<bytecode 0x477a0725>)(#<jupyter-repl-client jupyter-repl-client-5990c504> nil "incomplete" "")
  apply(#f(compiled-function (client req msg) #<bytecode 0x477a0725>) #<jupyter-repl-client jupyter-repl-client-5990c504> (nil "incomplete" ""))
  jupyter-handle-is-complete-reply(#<jupyter-repl-client jupyter-repl-client-5990c504> nil "incomplete" "")
  jupyter-repl-ret(nil)
  funcall-interactively(jupyter-repl-ret nil)
  call-interactively(jupyter-repl-ret nil nil)
  command-execute(jupyter-repl-ret)
```

Once it gets into this state, the only recovery that I've been able to figure out is to kill the REPL buffer and reconnect.  That's not ideal because the REPL history is specific to the buffer so it gets lost.

(FWIW, other times when the connection drops, I get a ZMQ websocket closed error instead, and restoring the connection to the server is enough to recover, unlike this case.  Because this happens only sporadicaly, I haven't been able to figure out what causes one or the other but I'd appreciate any advice on tracking that down).

This is installed from master (0a7055d) via Straight, emacs version `GNU Emacs 27.2 (build 1, aarch64-apple-darwin21.4.0, Carbon Version 165 AppKit 2113.4) of 2022-03-30`.

kleinschmidt:
patch from @pank :

https://github.com/nnicandro/emacs-jupyter/issues/395#issuecomment-1145976582

This does fix #395 for me.

### DIFF
--- a/jupyter-repl.el
+++ b/jupyter-repl.el
@@ -1197,7 +1197,8 @@ execute the current cell."
                                      "")))
               (jupyter-handle-is-complete-reply
                   jupyter-current-client
-                nil (if complete-p "complete" "incomplete") "")))
+                  nil (list :content (list :status (if complete-p "complete" "incomplete")
+                                           :indent "")))))
            (t
             (let ((res (jupyter-wait-until-received :is-complete-reply
                          (let ((jupyter-inhibit-handlers '(not :is-complete-reply)))


nnicandro:
Looks like I had already fixed this around 48eb141ea0665c6ca65c2473a3f85c3a35a0a946.

PR #423, opened 2022-12-04
Skip headers from jupyter kernelspec list.

ydl7:
On certain configurations, `juypter kernelspec list` has a header with irrelevant information. This causes an error when parsing JSON output and can simply be ignored.

Example header:

```
0.00s - Debugger warning: It seems that frozen modules are being used, which may
0.00s - make the debugger miss breakpoints. Please pass -Xfrozen_modules=off
0.00s - to python to disable frozen modules.
0.00s - Note: Debugging will proceed. Set PYDEVD_DISABLE_FILE_VALIDATION=1 to disable this validation.
```

### DIFF
--- a/jupyter-kernelspec.el
+++ b/jupyter-kernelspec.el
@@ -103,8 +103,10 @@ REFRESH."
           (or (and (not refresh) (jupyter-kernelspecs-cache-get host))
               (let ((specs
                      (plist-get
-                      (let ((json (or (jupyter-command "kernelspec" "list"
-                                                       "--json" "--log-level" "ERROR")
+                      (let ((json (or (let* ((jupyter-output (jupyter-command "kernelspec" "list"
+                                                                              "--json" "--log-level" "ERROR"))
+                                             (json-start (string-match-p (rx line-start "{") jupyter-output)))
+                                        (and json-start (substring jupyter-output json-start)))
                                       (error "\
 Can't obtain kernelspecs from jupyter shell command"))))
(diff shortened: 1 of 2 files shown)


nnicandro:
Do you still get those messages even when applying the fix in #418?

lesteve:
I get this warning with Python 3.11. The fix in #418 (adding `--log-level ERROR`) does not help:
```
❯ jupyter kernelspec list --json --log-level ERROR
0.00s - Debugger warning: It seems that frozen modules are being used, which may
0.00s - make the debugger miss breakpoints. Please pass -Xfrozen_modules=off
0.00s - to python to disable frozen modules.
0.00s - Note: Debugging will proceed. Set PYDEVD_DISABLE_FILE_VALIDATION=1 to disable this validation.
{
  "kernelspecs": {
    "python3": {
      "resource_dir": "/home/lesteve/.local/share/jupyter/kernels/python3",
      "spec": {
        "argv": [
          "python",
          "-m",
          "ipykernel_launcher",
          "-f",
          "{connection_file}"
        ],
        "env": {},
        "display_name": "Python 3 (ipykernel)",
        "language": "python",
        "interrupt_mode": "signal",
        "metadata": {
          "debugger": true
        }
      }
    }
  }
}
```

It seems to me that the best fix is to only consider stdout when executing jupyter commands as was proposed in #388, rather than adding more and more logic to discard unuseful info, which is what #418 did and and this PR proposes.

PR #425, opened 2022-12-21
ignore kernelspec warning output

GTrunSec:
Fixes: #418  #301
This PR will be a better solution to solve the parsing error for the kernel --json.  Just copied it from Ein project.


nnicandro:
Thanks for this but I would prefer to use #418. I would like all calls to the Jupyter programs to go through `jupyter-command` and I think it makes sense to fail if there is some error output. Although we should probably fail with a good message if there is some error happening when trying to parse the kernelspecs.

PR #429, opened 2023-01-28
Allow resolution of file extension when :file is specified

torfjelde:
This is a minor change which allows resolving the filename extension from the MIME type even if `:file` has been specified (and no specific extension was specified in `:file`).

This can be very useful when the MIME type of the output can change between evaluations but one still wants to use filenames other than the automatically generated one from `jupyter-org-image-file-name`.

One pain-point I repeatedly run is that I want to specify `:file`, but the file-extension used is not something I know ahead of time, e.g. if I'm plotting something in Julia, the resulting MIME type will be `png` if I've run `pyplot()` in some previous src-block, while it will be `svg` if I've run `gr()` in a previous src-block. Then it's useful for me to just specify the "base-name", say, `:file image/output` and have `jupyter-emacs` just add the extension based on the output.

A slightly more complex example: (with this branch) I like to add

```org
:file (f-join "assets" "outputs" (sha1 (plist-get (cadr (org-element-at-point)) :value)))
```

as buffer-wide header-args to my Julia blocks. This then generates the filenames from the *contents* of the block rather than the output. This becomes very useful when I then want to publish the resulting files on, say, Github: if I make a change to one src-block, which then alters one of the later figures, I can just re-run the code, get the new figure *which has the same filename as the previous run*, and then I get a diff that is just replacing this one binary file. In contrast, if I use the default provided by `jupyter-org-image-file-name` I now need to add the new file I just generated *and* remove the old file that is no longer referenced by my org-file.

One thing that might be worth changing about my current implementation: maybe it should be a opt-in feature? As in, the current behavior of `#master` stays but then if you do, say, `(setq jupyter-org-add-image-ext-when-needed t)` (which is `nil` by default) it will infer the file extension from the MIME type.

(Also, thank you so much for this awesome package! I've been using it daily for more than a year now, and really couldn't do without it :heart:)

### DIFF
--- a/jupyter-org-client.el
+++ b/jupyter-org-client.el
@@ -862,6 +862,18 @@ EXT is used as the extension."
 
 (defvar org-image-actual-width)
 
+(defun jupyter-org--add-image-ext-maybe (file &optional ext)
+  "Return a file name for FILE with EXT.
+If FILE is an image file with a specified extension, return FILE.
+If FILE is an image file without a specified extension and EXT is
+non-nil, return FILE with EXT as the extension.  Otherwise, return
+FILE."
+  (cond
+   ((null file) file)
+   ((image-type-from-file-name file) file)
+   ((not (null ext)) (concat file "." ext))
+   (t file)))
+
 (defun jupyter-org--image-result (mime content params &optional b64-encoded)
   "Return an org-element suitable for inserting an image.
 MIME is the image mimetype, CONTENT is a property list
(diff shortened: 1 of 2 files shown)


nnicandro:
Hey, thanks for the pull request and kind words about the project. I like the idea and think it is OK to have it  be the default behavior.

Could you add some information about this feature in the README, under the `Fixing the file name of images with the =:file= argument` section so that users can be aware of it.

jupyter-org-client.el
```diff
@@ -284,6 +284,8 @@ to."
         ;; TODO: Handle all of the different macro types for inline results, see
         ;; `org-babel-insert-result'.
         (setq data `(:text/plain ,(plist-get data :text/plain)))
+        ;; NOTE(tor): This is where the magic happens. Could potentially
```
nnicandro, line 287: This comment doesn't seem related to the pull request feature, at least I can't tell what it is referring to. Can you remove it or comment more how this would be related to the feature so that other people who read the code can understand its purpose.

torfjelde: Made an attempt at that now:) Sorry for the delay!

nnicandro:
Thanks.

PR #442, opened 2023-03-27
Fix jupyter-org--parse-latex-element

akirakyle:
Upstream org commit f01390c changed the behavior of org-inside-LaTeX-fragment-p which now checks for both latex fragments and environments

### DIFF
--- a/jupyter-org-client.el
+++ b/jupyter-org-client.el
@@ -1039,20 +1039,13 @@ parsed, wrap DATA in a minipage environment and return it."
   (with-temp-buffer
     (insert data)
     (goto-char (point-min))
-    ;; Try to determine if we are in an environment or fragment
-    (if (save-excursion
-          (forward-char 2)
-          (org-inside-LaTeX-fragment-p))
-        (org-element-latex-fragment-parser)
-      ;; If we are not in a fragment, try to parse an environment
-      (let ((env (ignore-errors
-                   (org-element-latex-environment-parser
-                    (point-max) nil))))
-        (if (eq (org-element-type env) 'latex-environment) env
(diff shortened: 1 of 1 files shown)


jupyter-org-client.el
```diff
@@ -1034,20 +1034,15 @@ parsed, wrap DATA in a minipage environment and return it."
...
-      (let ((env (ignore-errors
-                   (org-element-latex-environment-parser
-                    (point-max) nil))))
-        (if (eq (org-element-type env) 'latex-environment) env
-          ;; If all else fails, wrap DATA in a minipage
-          ;; environment
-          (jupyter-org-latex-environment
-           (concat "\
+    (let ((type (org-element-type (org-element-context))))
+      (cond ((eq type 'latex-fragment)
```
nnicandro, line 1038: Since the parser functions essentially return the same thing as `org-element-context` does why don't we return the value of `org-element-context` when `(memq type '(latex-fragment latex-environment))`?

akirakyle: Good point, I've updated the PR

akirakyle: PS: sorry for the slow response on this

akirakyle:
I opened another PR which adds org to emacs-jupyter's `Package-Requires` metadata field with version 9.1.6, however since this fixes an upstream change made later, that might need to bumped up to 9.6.1 I think. I'm not sure how you want to handle this exactly since emacs 28 comes with builtin org 9.5, and emacs 29 isn't out yet which I think will have 9.6 in it.

nnicandro:
For this pull request your solution will work regardless of Org version since it uses `org-element-context`.

PR #467, opened 2023-06-16
Add org to Package-Requires

akirakyle:
Trying to use the elpaca package manager currently gives bytecomp errors of "Error: Org version mismatch." when trying to use a non-builtin org version. The elpaca dependency manager doesn't load the newer version of org in the bytecomp process since it isn't declared in the Package-Requires section.

I wasn't totally sure what minimum org version to specify, so I went with the org version that shipped with emacs 26.


nnicandro:
Thanks.

PR #473, opened 2023-06-24
Fix for #470

### LINKED ISSUE #470: Deferred aliases causing issues
I believe https://github.com/emacs-jupyter/jupyter/commit/11ace8c805e4f866005a77891b29c4fe036c57c8 is causing some issuess. AFAIK this effectively means that we need to execute a block before you can even start editing the block, since Emacs will complain that functions such as `org-babel-edit-prep:jupyter-julia` don't exist :confused: 

I'm on the run atm, so will get back to this properly later, but figured I should at least put it here asap.

torfjelde:
This PR fixes #470  by adding the alias-definitions "hook" to run upon execution of different `org-edit-*` methods too.

An alternative might be to just add the device to a method a bit "deeper" in the call-stack, e.g. `org-babel-get-src-block-info`, which, I would assume would need to be queried for most of the different methods before hitting any of the potentially aliased methods.

### DIFF
--- a/ob-jupyter.el
+++ b/ob-jupyter.el
@@ -43,6 +43,8 @@
 (declare-function org-element-property "org-element" (property element))
 (declare-function org-element-context "org-element" (&optional element))
 (declare-function org-babel-execute-src-block "ob-core" (&optional arg info params executor-type))
+(declare-function org-edit-special "org" (&optional arg))
+(declare-function org-edit-src-block "org-src" (&optional code edit-buffer-name))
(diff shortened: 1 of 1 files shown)


nnicandro:
I'd rather make the aliases in `org-mode-hook`. This way there is no need to worry about any user interaction to generate the aliases. Any interaction with an Org buffer already implies that `org-mode` gets called.

torfjelde:
That crossed my mind, but it wasn't completely obvious to me that this would be suffficient for the conda-env setup :shrug: But I do like that approach better, if it's indeed sufficient :+1:

Issue #446, opened 2023-04-23
`jupyter-run-repl` stopped working


EmmanuelCharpentier:
Setup : 

- `This is GNU Emacs 28.2 (build 1, x86_64-pc-linux-gnu, GTK+ Version 3.24.37, cairo version 1.16.0) of 2023-03-14, modified by Debian`
- from `packages-list-package` : `jupyter 20230214.215` (installed from MELPA).

In a fresh `emacs` session, `M-x jupyter-run-repl` fails with `Invalid number format: 5`. In the `*Messages*` buffer : 
```
Loading /etc/emacs/site-start.d/00debian.el (source)...done
Loading /etc/emacs/site-start.d/50asymptote.el (source)...done
Loading /etc/emacs/site-start.d/50autoconf.el (source)...done
Loading /etc/emacs/site-start.d/50dictionaries-common.el (source)...
Loading debian-ispell (native compiled elisp)...
Loading /var/cache/dictionaries-common/emacsen-ispell-default.el (source)...done
Loading debian-ispell (native compiled elisp)...done
Loading /var/cache/dictionaries-common/emacsen-ispell-dicts.el (source)...done
Loading /etc/emacs/site-start.d/50dictionaries-common.el (source)...done
Loading /etc/emacs/site-start.d/50gnuserv.el (source)...done
Loading /etc/emacs/site-start.d/50latex-cjk-common.el (source)...
Loading /usr/share/emacs/site-lisp/latex-cjk-common/cjk-enc.el (source)...done
Loading /etc/emacs/site-start.d/50latex-cjk-common.el (source)...done
Loading /etc/emacs/site-start.d/50latex-cjk-thai.el (source)...done
Loading /etc/emacs/site-start.d/50latexmk.el (source)...done
Loading /etc/emacs/site-start.d/50lilypond-data.el (source)...done
Loading /etc/emacs/site-start.d/50noweb.el (source)...done
Loading /etc/emacs/site-start.d/50texlive-lang-english.el (source)...done
Loading ispell (native compiled elisp)...done
For information about GNU Emacs and the GNU system, type C-h C-a.
json-read: Invalid number format: 5
```

This setup (more or less) worked two days ago with the `ir` kernel (minor issue wit vectors display format) and was nonfunctional with the Wolfram Engine kernel (no output display, see #223).

Results of running `M-! jupyter kernelspec list` : 
```
Available kernels:
[...]

PR #483, opened 2023-07-24
Parse json from kernelspec being printed with warnings

ahmed-shariff:
Potential fix for #446 
`jupyter kernelspect` seems to be spitting a few warnings before the json. This PR strips out that and parses only the json. Also, prints out a warning if there are any warnings.


ahmed-shariff:
shoot, this might be a duplicate of #423

Issue #492, opened 2023-09-04
org-babel-after-execute-hook not executed in context when called from jupyter-handle-execute-reply


ahmed-shariff:
I have a hook added to the `org-babel-after-execute-hook` which is expected to run with point at the source block. But I get the following error:
```emacs-lisp
Jupyter: I/O subscriber error: (error "‘org-element-at-point’ cannot be used in non-Org buffer #<buffer  *zmq*-120077> (special-mode)")
```

The above happens with `jupyter-python` in:
- emacs 29.1
- org mode version 9.7-pre (release_9.6.8-745-gdbf415)
- emacs-jupyter 1baabc8

Based on my exploration, the issue seems to be happening in this method:
https://github.com/emacs-jupyter/jupyter/blob/1baabc8f2d0631f8f4308400f9a53f1aadd58d14/jupyter-org-client.el#L340-L350

When the `org-babel-after-execute-hook` is called, seems like it is not being called from the context of the source block. As a workaround I have the following extension method:

```emacs-lisp
  (cl-defmethod jupyter-handle-execute-reply :around ((_client jupyter-org-client) (req jupyter-org-request) msg)
    (org-with-point-at (jupyter-org-request-marker req)
      (cl-call-next-method)))
```

nnicandro:
Thanks, I fixed this in 0480c47.

PR #497, opened 2023-10-08
Fix jupyter-org--parse-latex-element again

akirakyle:
Upstream org commit 4fcdcb4 changed the behavior of `org-element-context` which now uses `org-element-deferred-create` which is incompatible with inserting elements created from with-temp-buffer outside its scope.

Note that the first error I ran into which led to this PR was actually
```
Jupyter: I/O subscriber error: (error "‘org-element-at-point’ cannot be used in non-Org buffer #<buffer  *temp*> (fundamental-mode)")
```
which at first I thought was related to #492 which is due to upstream org commit f135954, however I realized that that issued seemed to be fixed in master already and simply adding a call to `(org-mode)` in `jupyter-org--parse-latex-element` fixed that error but exposed this deeper one which I had missed since I upgraded from a commit prior to 4fcdcb4 past commit f135954.

### DIFF
--- a/jupyter-org-client.el
+++ b/jupyter-org-client.el
@@ -811,14 +811,6 @@ Otherwise, return VALUE formated as a fixed-width `org-element'."
    (t
     (org-element-create 'fixed-width (list :value (format "%S" value))))))
 
-(defun jupyter-org-latex-fragment (value)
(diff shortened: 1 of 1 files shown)


akirakyle:
Closing for #514 since I accidentally made this PR based on my fork's master branch.

PR #501, opened 2023-10-16
ob-jupyter.el: Delay `org-babel-jupyter-make-local-aliases` hook.

jamieforth:
Ensuring this function is called after `jupyter-org-interaction-mode` in `org-mode-hook` to ensure correct fontification of errors (i.e. handling ansi-color escape characters).


nnicandro:
For reference, [here](https://github.com/emacs-jupyter/jupyter/issues/476#issuecomment-1659428030) is the comment mentioning why this change fixes the issue of ANSI fontification not working on first opening of an Org file.

PR #510, opened 2023-11-05
request raw history

malb:
If raw is nil then with SageMath kernels the history we're getting is `Integer(1) + Integer(2)` if `1+2` was issued as a command before, i.e. SageMath does some heavy processing of inputs.

### DIFF
--- a/jupyter-repl.el
+++ b/jupyter-repl.el
@@ -1681,7 +1681,7 @@ Return the buffer switched to."
       (jupyter-sent
        (jupyter-history-request
         :n jupyter-repl-history-maximum-length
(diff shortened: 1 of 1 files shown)


malb:
There might be a reason I'm missing why `:raw nil` was set, but I figured a PR that you can accept and decline is quicker than an issue first.

nnicandro:
I don't think there was a reason why `:raw nil` was added, atleast not what I can gather from the commit history. I don't think setting `:raw t` would hurt anything, it is a better setting anyways since that is what the user types in at the REPL.

Issue #486, opened 2023-07-28
Executing blocks with malformed results section erases org file contents


pati-ni:
In some scenarios (does not happen often) the executing a src block erases the appended commend of the org-mode file. The result is that everything at least until the next src block gets erased.

My guess is that in the #+RESULTS: section the :RESULTS: are not properly wrapped with the :END: block. Thus, re-executing the src block erases everything up until the next :END: block. If lucky and executed the next block it will stop in :END: of that block. If not, most likely it will erase everything. Usually this can be reverted, if I realize described behavior with an undo.

But this can lead to some nasty situations. Is there a way to introduce a fail-safe for this?

ed9w2in6:
I face a similar symptom *regularly* but only on commands that fails and triggers error / warnings.
In which `org-mode` will complain, here are an excerpt of my `*Warnings*` buffer

```
 ■  Warning (org-element-cache): org-element--cache: Unregistered buffer modifications detected (312536 != 312301). Resetting.
If this warning appears regularly, please report the warning text to Org mode mailing list (M-x org-submit-bug-report).
The buffer is: scratch.org
 Current command: nil
 Backtrace:
nil
```

So a good starting point on any investigation would be:
- how `emacs-jupyter` put error message (`[goto-error]`?) into org buffer
- what `org-element-cache` expects buffer are edited.

Sorry that I do not have time right now to look further into this but I shall investigate more on it later since it affects me a lot.
I used to just ignore it but lately I found that huge sections of my org file got deleted away without me noticing.
Undo also won't work but luckily I had made a backup few weeks ago.

nnicandro:
Could you give an example where Emacs-Jupyter causes the `:END:` line to be missing from the results of a source block? There shouldn't be any such cases happening, if there are let me know.

It would be great if you could provide a minimum working example Org file with example source blocks that can reproduce your problem.  I am aware of the scenario that you mention, but evaluating source blocks with Emacs-Jupyter should always create valid Org documents.

akirakyle:
I have also been seeing these issues but haven't had a chance to debug further and create a minimum working example. I suspect the org-element--cache issues may be related to to the way jupyter handles ansi colors as I still see inconsistent behavior for error outputs that have escape characters to color them.

ed9w2in6:
@nnicandro 
For me, I am not sure if `:END:` line is missing all the time when erasure of contents happen.
I am sure that:
- if the code fails and triggers error / warnings, content MIGHT be erased.
[...]

PR #515, opened 2023-11-24
Fix org-element-cache issue with ansi colors

akirakyle:
I was getting really annoyed with all the `org-element-cache` warnings using ob-jupyter with julia which are described in #486 so I tried to figure out what was causing them. I'm not sure if this is the right way to fix it, or if it may also be caused by other things in emcas-jupyter, but so far I haven't been seeing the issue with this patch for my workflows.

### DIFF
--- a/jupyter-mime.el
+++ b/jupyter-mime.el
@@ -205,23 +205,14 @@ value of t is added to render the escape sequences invisible.
 Also, the `ansi-color-apply-face-function' is hard-coded to a
 custom function that prepends to the face property of the text
 and also sets the FACE-PROP to the prepended face, if FACE-PROP
-is nil it defaults to `font-lock-face'.
-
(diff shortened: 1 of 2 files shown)


MasonProtter:
I tried this branch to see if if fixed #522, and while I no longer get the warnings, the ansi colours are no longer displayed:
```
#+begin_src jupyter-julia :session foo
@info "hi"
#+end_src

#+RESULTS:
: [36m[1m[ [22m[39m[36m[1mInfo: [22m[39mhi
```

akirakyle:
Can you try `revert-buffer` to see if the ansi escape sequences get rendered? I've encountered this here but haven't had time to debug it.

MasonProtter:
Oh interesting, yeah a `revert-buffer` did work!

PR #555, opened 2024-08-24
Read websocket frame until completed

kwrooijen:
## Problem

I was trying out emacs-jupyter to connect to [Paperspace Gradient](https://www.paperspace.com) using their token authenticated [remote kernel](https://docs.digitalocean.com/products/paperspace/notebooks/how-to/connect-remote-kernel/). I noticed that some code executions didn't output anything (e.g. `.describe()`). 

## Solution

It turns out that the current `:on-message` handler doesn't properly handle partial messages. If `websocket-frame-completep` returns `nil` we should wait for the next frame and accumulate the message.

We also add `continuation` `websocket-frame-opcode` as a valid string response. I'm not sure what impact this will have since `continuation` isn't actually a type like `text` or `binary`

## Note

For those interested in connecting to Paperspace Gradient; you have to connect WITHOUT using a port. AFAIK emacs-jupyter won't let you set the `port` to `nil` _in org_, so you have to manually override `jupyter-tramp-url-from-file-name` and set `port` to `nil`
 https://github.com/emacs-jupyter/jupyter/blob/f97f4b5d8c83e0b901020f835183dde8a2bf649e/jupyter-tramp.el#L282-L289

### DIFF
--- a/jupyter-server-kernel.el
+++ b/jupyter-server-kernel.el
@@ -193,6 +193,19 @@ Call the next method if ARGS does not contain :server."
     (jupyter-run-with-io pub
       (jupyter-publish 'reauthenticate))))
 
+(defvar jupyter-server-kernel-frame-acc "")
+
+(defun jupyter-server-kernel-parse-frame (frame)
+  "Accumulate the WebSocket frame data and return the full response when complete.
+Return nil if the frame is not yet complete."
+  (let ((payload (websocket-frame-payload frame)))
+    (setq jupyter-server-kernel-frame-acc (concat jupyter-server-kernel-frame-acc payload))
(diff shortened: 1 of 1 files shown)


nnicandro:
Thanks for this. I think having `continuation` like that will be fine based on how you've written it and how Websocket sends partial messages (based on the RFC).

Couldn't you just specify the default HTTP or HTTPS ports in the session string in Org, like `/jpy:host:80` or `/jpys:host:443`, so that you don't have to override that function?

kwrooijen:
I don't exactly recall what the problem was in my note. I believe the issue was that it just didn't work if you specified any port (80 or 443). I'll get back to you once I have time (with a new issue).

Thanks for merging!

PR #563, opened 2024-11-18
Fix control characters appearing in connection file.

ydl7:
Sometimes control characters appear in the connection file name, this is a trivial fix to modify the regular expression to avoid this. One could also improve the expression to pick out the JSON filetype probably, but I wanted to keep the change minor.

### DIFF
--- a/jupyter-env.el
+++ b/jupyter-env.el
@@ -165,7 +165,10 @@ The session can be used to write a connection file, see
                (error "`jupyter kernel` failed to show connection file path"))
         (and (process-live-p process)
              (goto-char (point-min))
-             (re-search-forward "Connection file: \\(.+\\)\n" nil t)))
+             (re-search-forward (rx "Connection file: "
+                                    (group (+ (not cntrl)))
+                                    (* whitespace) line-end)
+                                nil t)))
       (let* ((conn-file (concat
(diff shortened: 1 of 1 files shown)


nnicandro:
The group will match some whitespace at the end of the grouping since a space is included as a non-control character. It would be safer to call `string-trim-right` on the result of the grouping to account for this. Although the original regex doesn't account for this either.

Also, can't a control character be a part of a valid file name?  I just tried it on Linux and I was able to create a file with a tab character in it like `fo\to`.

ydl7:
Hi @nnicandro,

That's all correct. In principle there can be several other issues as well.

There are two more "aggressive" changes, you can tell me which you prefer and I can modify the PR:
1. Match so that there is a non-control, non-whitespace character at the end of the filename (but allowing control characters up until the very end).
2. Match the actual `.json' filetype of the connection file.

Let me know,

Yuri

nnicandro:
The second option seems better. There probably aren't any cases where the connection file would not have a JSON extension and the intention is clearer that we are looking for a JSON file specifically as the rest of the code assumes a file with JSON contents anyways.

ydl7:
@nnicandro updated to only match JSON files.

PR #572, opened 2025-02-14
Make org keybindings and integration optional

isamert:
This PR introduces two variables to control the integration with org:

- jupyter-org-want-keybinding
- jupyter-org-want-integration

which are true by default to keep the old behavior. 

I believe org interaction should be optional because any breakage in jupyter package causes issues with org and that is a very frustrating experience for heavy org users. Now with these variables, user can disable the integration for the time being or they can simply turn it on manually on demand.

### DIFF
--- a/jupyter-org-client.el
+++ b/jupyter-org-client.el
@@ -123,6 +123,11 @@ See also the docstring of `org-image-actual-width' for more details."
   :group 'ob-jupyter
   :type 'boolean)
 
+(defcustom jupyter-org-want-keybinding t
+  "Whether to enable contextual keybindings."
+  :group 'ob-jupyter
+  :type 'boolean)
+
 (defconst jupyter-org-mime-types '(:text/org
                                    ;; Prioritize images over html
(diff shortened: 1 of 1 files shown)


ob-jupyter.el
```diff
@@ -834,10 +834,11 @@ mapped to their appropriate minted language in
 (defun org-babel-jupyter-make-local-aliases ()
   (let ((default-directory user-emacs-directory))
     (org-babel-jupyter-aliases-from-kernelspecs)))
-(add-hook 'org-mode-hook #'org-babel-jupyter-make-local-aliases 10)
 
-(add-hook 'org-export-before-processing-functions #'org-babel-jupyter-setup-export)
-(add-hook 'org-export-before-parsing-functions  #'org-babel-jupyter-strip-ansi-escapes)
+(when jupyter-org-want-integration
```
nnicandro: Why should this be here? There would be no way to recognize Jupyter source blocks in the buffer if we don't make the aliases at some point and the other hooks only apply for Jupyter specific source blocks as well.

PR #573, opened 2025-02-18
Add a GIF to illustrate the REPL workflow

### LINKED ISSUE #535: Increase visibility of the package
Thanks for the package!

It strikes me that `emacs-jupyter` seems relatively unknown in the wider emacs/python communities, despite providing the best interactive python workflow of any editor that I have worked with (having spent significant time on each of VSCode, PyCharm, JupyterLab, and Neovim). The ability to disentangle python source code and python REPL sessions is a big improvement over the standard jupyter notebook workflow. But `emacs-jupyter` is unique in that it provides a python REPL that behaves much like a regular buffer. Arguably this provides a superior experience to any other REPL; one consistent set of keybindings everywhere! The closest alternative would be Neovim via some of the below plugins, however these all fall short because they run the python REPLs using the likes of vterm, which unfortunately has its own set of modes/keybindings (not to mention graphics integration becomes a huge headache).

Given the merits of `emacs-jupyter`, I think it would benefit from being better publicized. In particular, I think a few screenshots or even short videos of a typical python script + REPL workflow would go a long way, and hopefully increase adoption. For example, the following vim packages do a good job of making this functionality clear right away:
- https://github.com/jpalardy/vim-slime
- https://github.com/sillybun/vim-repl
- https://github.com/luk400/vim-jukit

I think adding something similar to the README could really catch the eye of users that are interested in this type of interactive python workflow, who might otherwise gloss over this awesome package and get lost in all the other org-related stuff that's also documented in the README (on a related note, [Doom emacs](https://github.com/doomemacs/doomemacs) only includes `emacs-jupyter` as an extension of its `org` module, but not as an extension of its `python` module, perhaps because there is indeed a lack of awareness of the awesome standalone REPL that `emacs-jupyter` provides).

egr95:
Closes #535


junghan0611:
This is the workflow I was looking for, thank you.

junghan0611:
How did you set dotfiles up emacs-jupyter and code-cells?

egr95:
Glad you like it!

My config uses the awesome [doom emacs](https://github.com/doomemacs/doomemacs), which I highly recommend (note that emacs-jupyter is shipped as part of the `org` module in doom, so you can choose between enabling that and installing manually). Other than that, I like to use `(setq jupyter-repl-echo-eval-p t)`, which conveniently echoes the code you have run in the REPL, and I also define the following function for sending code chunks from my python files to the REPL (which I borrowed from [Hank Greenburg](https://gitlab.com/greenbeast/dotfiles/-/blob/work-setup/init.org)):
```
(defun gm/jupyter-eval-region (beg end)
  "Evaluate the region between BEG and END."
  (interactive "r")
  (let* ((string (buffer-substring beg end))
         (string (replace-regexp-in-string "\\`[\n]*" "" string)) ; Remove leading empty lines
         (indent-length (string-match "[^ \t]" string)) ; Find indent length of the first line
         (unindented-string (replace-regexp-in-string (format "^%s" (make-string indent-length ?\ ))
                                                      "" string t t))) ; Remove exactly that amount of indentation
    (jupyter-eval-string unindented-string)))
(add-to-list 'code-cells-eval-region-commands '(jupyter-repl-interaction-mode . gm/jupyter-eval-region))
```

Issue #313, opened 2021-02-22
jupyter-load-file -- "/path file/with spaces"


SciBourne:
Hi, sorry for my English :)
When I try to upload a file with spaces in path, I get a Traceback:
```

OSError: File `'/very/long'` not found.

During handling of the above exception, another exception occurred:

Exception                                 Traceback (most recent call last)
<ipython-input-44-79d3b1904738> in <module>
----> 1 get_ipython().run_line_magic('run', '/vey/long path/to/file.py')

```

I use a temporary hack for this case: 

```
(defun jupyter-load-file-test ()
  (interactive)
  (jupyter-eval-string
   (jupyter-load-file-code (concat "'" buffer-file-name "'"))))

(global-set-key   (kbd "C-c f")  'jupyter-load-file-test)

```

I don’t know, maybe it will be useful ...

nnicandro:
You can add the quotes to the argument of the function below

https://github.com/nnicandro/emacs-jupyter/blob/6ce8d01e3a550a3268b415bf9d9b635d4dba5940/jupyter-python.el#L77-L78

and submit a pull request.

But I think you should use double quotes since single quotes are not supported on Windows according to the last point [here](https://ipython.readthedocs.io/en/stable/interactive/magics.html#magic-run).

Issue #579, opened 2025-03-20
jupyter-load-file doesn't quote filenames


moneylobster:
Windows allows spaces in filenames. On Windows, sending a `jupyter-load-file` command to a Python kernel will result a query like `%run C:/Hello There/test.py` to be evaluated, which causes problems because a python script called `C:/Hello` doesn't exist. Quoting the filename in `jupyter-load-file-code` seems to fix the problem:
```elisp
(cl-defmethod jupyter-load-file-code (file &context (jupyter-lang python))
  (concat "%run \"" file "\"")) ;; instead of (concat "%run " file)
```

moneylobster:
Just found out that this is a duplicate of #313 , sending a PR with the relevant change.

PR #580, opened 2025-03-20
Quote filename when calling jupyter-load-file

moneylobster:
Not quoting filenames causes the path to be misinterpreted if it contains spaces. Fixes issues #313 and #579 .


nnicandro:
You have commits from some other repository. Can you pull down the most recent changes for this repo and just make the one change you want.

PR #582, opened 2025-04-05
Remove nested list in jupyter-org--src-block-cache initialisation

jamieforth:
`(jupyter-org--set-src-block-cache)` throws `error: (wrong-type-argument markerp nil)` because ` jupyter-org--src-block-cache` is initialised as a nested list.


EduardoVernier:
I'm getting this same error, thank you for working on this.

PR #591, opened 2025-06-28
Fix: set org src block cache (fix autocomplete)

dalanicolai:
The completion-at-point functionality is broken because the org src-block-cache is set to a nested list, instead of just a list. Therefore, the pcase-let destructuring in the
jupyter-org--src-block-cache does not work correctly, and fails to bind the cached markers. 'Unnesting' the cache-list fixes the set-cache and completion functionality.

I did not 'investigate' how the what the src-cache is supposed to look like, but it makes sense to me that it should not be a list of lists (as there should probably be only one 'cache' per src-block).


dalanicolai:
Oh... only now I see that #582 already fixes this (just it is not a very elaborate issue description)

